PR #1, opened 2018-12-08
Update README.rst

amancevice:
I think you mean "sensitive" here. No production system I've ever worked with has sensible data in it ;)


Delgan:
Oh, that's kind of you to have fixed this typo, thank you!

PR #16, opened 2018-12-14
Update setup.py

Unviray:


Delgan:
Hey @Unviray, could you please add some context about why your think these changes are needed?

For example, the [`setup.py` of `requests`](https://github.com/requests/requests/blob/master/setup.py) doesn't use `if __name__ == "__main__":` and it never seems to have been a problem.

Also, not using f-strings is intended, as `loguru` aims to be compatible with Python 3.5 in the near future.

PR #19, opened 2018-12-17
Removes unused imports

sobolevn:


Delgan:
Oops, memories of my first attempts at implementing Loguru in C... Thanks for the cleanup!

PR #23, opened 2018-12-24
Typo

ss18:
occured -> occurred

Found with: https://github.com/ss18/grep-typos


Delgan:
Nice library, thanks for the correction. 👍

PR #24, opened 2018-12-24
Correct a grammar mistake

tyong920:
This library is awesome. I can't help myself reading the code. Thanks for your excellent work!


Delgan:
Thanks for both the kind words and the fix! 😃

PR #34, opened 2019-01-16
Fixes "MMMM" and "MMM".

nasyxx:
There is no need to `-1` here.

> ### calendar.month_name
> An array that represents the months of the year in the current locale. This follows normal convention of January being month number 1, so it has a length of 13 and month_name[0] is the empty string.

> ### calendar.month_abbr
> An array that represents the abbreviated months of the year in the current locale. This follows normal convention of January being month number 1, so it has a length of 13 and month_abbr[0] is the empty string.


Delgan:
Thanks a lot for the fix, it was a very unfortunate bug.

PR #35, opened 2019-01-16
Fixes a typo.

nasyxx:
`{"sink": sys.stdout, format="{time} - {message}"}`

to

`{"sink": sys.stdout, "format": "{time} - {message}"}`


Delgan:
Thanks for having spotted that! 👍

PR #40, opened 2019-01-20
Travis: codecov: pass Python version as flag

blueyed:


Delgan:
@blueyed Why do you think passing the python version to coverage command is useful?

I don't know much about `codecov`, so I don't know what the `--flags` is used for.

blueyed:
@Delgan 
Codecov uses this to group coverage reports, i.e. you can see which code is covered by which Python version.  There can be multiple flags, e.g. also "linux" and "windows" could be flags later.

blueyed:
See https://codecov.io/gh/Delgan/loguru/compare/49a34526ec0491b1d2338b799187463453688738...7d2b5361a4c08f0a215e0ee53707b880abc65e07.

PR #55, opened 2019-02-17
Fix minor typos in _logger.py's comments

needoptimism:
Hello!
I encountered this project, while being annoyed by boiler plating part of standard python logging library.
And, here is a fix for some minor typos!


Delgan:
Thanks a lot @needoptimism for these corrections! 😁

PR #67, opened 2019-03-15
Include tests in PyPI tarball

dotlambda:
This is useful for downstream distributions to test if the package works correctly with their build recipe. Also, it provides a way to test if the dependencies as they are packaged by the distribution are compatible.


Delgan:
Great, thanks for the improvement! 👍

PR #85, opened 2019-04-09
tests/test_formatting: use __file__

blueyed:
It could be used in more places there, but I think that is fine for now.

I've noticed this when copying the file to another name for a new test
being added there (to compare against master).

### DIFF
--- a/tests/test_formatting.py
+++ b/tests/test_formatting.py
@@ -18,7 +18,7 @@ from ansimarkup import AnsiMarkupError
         ("{level.icon}", lambda r: r == "🐞"),
         ("{file}", lambda r: r == "test_formatting.py"),
         ("{file.name}", lambda r: r == "test_formatting.py"),
-        ("{file.path}", lambda r: re.fullmatch(r".*tests[/\\]test_formatting.py", r)),
+        ("{file.path}", lambda r: r == __file__),
         ("{function}", lambda r: r == "test_log_formatters"),
         ("{module}", lambda r: r == "test_formatting"),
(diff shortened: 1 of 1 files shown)


Delgan:
Nice idea. However, this seems to fail on Windows. `{file.path}` is in lowercase, while `__file__` preserves the path case.

```
c:\users\delgan\desktop\loguru\tests\test_formatting.py
C:\Users\delgan\Desktop\loguru\tests\test_formatting.py
```

blueyed:
Should `{file.path}` get fixed then maybe?
Otherwise feel free to just close this.

Delgan:
> Should `{file.path}` get fixed then maybe?

Yep, that's definitely a bug!
The behavior differs from the standard ``logging``. I can't remember why I used `normcase()` (it's there since the first commit). Maybe did I mean `normpath()`?

About the proposition of replacing `tests/test_formatting.py` with `__file__`, I read that [in the Python source code](https://github.com/python/cpython/blob/a9bd8925c7fa50dd3cfab125b824ec192133ef49/Lib/logging/__init__.py#L171-L174):

> frozen modules don't always have \_\_file\_\_ set

Although I have trouble imagining Loguru's tests being executed this way, maybe it's best to avoid using \_\_file\_\_to avoid this bug by principle?

Delgan:
@blueyed Sorry for the delay before merging this, thanks for the improvement!

Also, I removed the `normcase()` for `path` in https://github.com/Delgan/loguru/commit/786151ca22ea63e2426bb6464a971455a2664a9c .

Issue #59, opened 2019-02-22
Pytest's caplog fixture doesn't seem to work


dougthor42:
# Summary

[Pytest's `caplog` fixture](https://docs.pytest.org/en/latest/logging.html#caplog-fixture) is a critical part of testing. I'd love to move to loguru, but loguru doesn't seem to work with `caplog`.

I'm not sure if this is user error (perhaps it's documented somewhere? I haven't been able to find it.), if it is some design oversight/choice, or if the problem is actually on pytest's end.

## Expected Result

Users should be able to use `loguru` as a drop-in replacement for the stdlib `logging` package and have tests that use the `caplog` fixture still work.

## Actual Result

Drop-in replacement causes tests that use the `caplog` pytest fixture to fail.

# Steps to Reproduce

## Base test file
```python
# test_demo.py
import pytest
import logging
logger = logging.getLogger()
logger.addHandler(logging.StreamHandler())
# from loguru import logger

def some_func(a, b):
    if a < 1:
        logger.warning("Oh no!")
    return a + b

def test_some_func_logs_warning(caplog):
    assert some_func(-1, 2) == 1
    assert "Oh no!" in caplog.text

if __name__ == "__main__":
    some_func(-1, 1)
    print("end")
```

## Without Loguru:

```shell
$ python test_demo.py
Oh no!
end
(.venv) Previous Dir: /home/dthor
09:59:56 dthor@Thorium /home/dthor/temp/loguru
$ pytest
========================== test session starts ==========================
platform linux -- Python 3.6.7, pytest-4.3.0, py-1.8.0, pluggy-0.8.1
rootdir: /home/dthor/temp/loguru, inifile:
collected 1 item

test_demo.py .                                                    [100%]

======================= 1 passed in 0.03 seconds ========================
```

## With Loguru:

Adjust `test_demo.py` by commenting out stdlib `logging` and uncommenting `loguru`:

```python
...
# import logging
# logger = logging.getLogger()
# logger.addHandler(logging.StreamHandler())
from loguru import logger
...
```

```shell
$ python test_demo.py
2019-02-22 10:02:35.551 | WARNING  | __main__:some_func:9 - Oh no!
end
(.venv) Previous Dir: /home/dthor
10:02:35 dthor@Thorium /home/dthor/temp/loguru
$ pytest
========================== test session starts ==========================
platform linux -- Python 3.6.7, pytest-4.3.0, py-1.8.0, pluggy-0.8.1
[...]

PR #86, opened 2019-04-09
[RFC] Allow to configure exception formatter

blueyed:
TODO:

- [ ] base class / interface for formatters?
- [ ] I've moved code out of the `with self._lock` block, but I think that is OK (initially added there in https://github.com/blueyed/loguru/commit/04aba79e6b0175eef075a0bb3776a90876da3332#diff-4446c03371d87929a4cefc5c88a52f36R764)

Came across this when looking at https://github.com/Delgan/loguru/issues/59, i.e. when you want to be compatible to `logging`, no exceptions should be added also.

### DIFF
--- a/loguru/_handler.py
+++ b/loguru/_handler.py
@@ -93,12 +93,12 @@ class Handler:
 
             formatter_record = record.copy()
 
-            if not record["exception"]:
-                error = ""
-            else:
+            if record["exception"] and self._exception_formatter:
                 type_, value, tb = record["exception"]
                 lines = self._exception_formatter.format_exception(type_, value, tb)
                 error = "".join(lines)
+            else:
+                error = ""
 
             formatter_record["exception"] = error
 
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -743,6 +743,23 @@ class Logger:
                 "Invalid level value, it should be a positive integer, not: %d" % levelno
             )
 
+        try:
+            encoding = sink.encoding
+        except AttributeError:
+            encoding = None
+
+        if not encoding:
+            encoding = "ascii"
+
+        if exception_formatter is UNSET:
+            exception_formatter = ExceptionFormatter(
+                colorize=colorize,
+                encoding=encoding,
+                diagnose=diagnose,
+                backtrace=backtrace,
+                hidden_frames_filename=self.catch.__code__.co_filename,
+            )
+
         if isinstance(format, str):
             formatter = format + "\n{exception}"
             is_formatter_dynamic = False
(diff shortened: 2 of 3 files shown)


Delgan:
I'm not sure to follow what is wrong with #59, is there a problem with the Pytest workaround provided? Some exceptions that are displayed while they should not?

Thanks for the suggestion but I don't want `exception_formatter` formatter to be part of the API. I don't expect custom exception formatting to be a very common use case, and it's currently already possible with just a few lines of code:

```python
def formatter(record):
    fmt = "{time} {name} - {message}"
    if record["exception"] is not None:
        record["extra"]["formatted_exception"] = my_exception_formatter(record["exception"])
        fmt += " {extra[formatted_exception]}\n"
    return fmt

logger.add(sys.stderr, format=formatter)
```

blueyed:
> I'm not sure to follow what is wrong with #59, is there a problem with the Pytest workaround provided? Some exceptions that are displayed while they should not?

It is not really compatible from what I can tell, since exceptions are added to the message - this at least does not really work when migrating from logging to loguru.  (However I might have understood the intention there)

As for this PR I've just thought that processing could be skipped in general if not required for example.

> I don't want exception_formatter formatter to be part of the API

I see.. then I will look into removing this from here, IIRC there are other useful bits in it.
My main motivation was to not have a custom one, but allow to disable it.

Your example code looks like it would do the job, but is a bit more involved (but fine).
It would still instantiate the internal formatter though, wouldn't it?

Delgan:
> It is not really compatible from what I can tell, since exceptions are added to the message - this at least does not really work when migrating from logging to loguru. (However I might have understood the intention there)

Ok, yes, I see, sorry. This is because  the formatter in `logger.add(sink, format="{message}")` is not simply `"{message}"` but actually `"{message}\n{exception}"`, is that what you mean?

So, when `logging.makeRecord()` is called to create a record compatible with standard ``logging``, the `.msg` attribute of the resulting `LogRecord` object will potentially contain the formatted exception, while it shouldn't.  Hence the [`exc_text = "\n"`](https://github.com/Delgan/loguru/blob/6571879c37904e3a18567e694d70651c6886b860/loguru/_logger.py#L734) hack that I added to prevent the exception to be displayed twice.
I'm starting remembering all these nasty things I have done... :)

> My main motivation was to not have a custom one, but allow to disable it.
>
> Your example code looks like it would do the job, but is a bit more involved (but fine).

The `"\n{exception}"` is automatically added for convenience, but there is actually a way to disable it by using a formatter function: `logger.add(sink, format=lambda _: "{message}\n")`. 

I think this would produce the same result that your proposition.

However, this implies that `logger.exception("Error")` would not propagate the traceback. Actually, the `LogRecord.msg` contains the log message formatted by Loguru, whether it is `"{message}"`, `"{message}\n{exception}"`, `"{time} {function} {message}"`, etc. That's way, the user can choose the way it prefer to format the propagated message (through standard `setFormatter()` or Loguru `format`).

> It would still instantiate the internal formatter though, wouldn't it?

Yes, but I'm not worried about that. The exception formatter is instantiated only once, while adding the handler, it's not a problem. When calling `logger.exception()`, the exception will be formatted but not added to the message, though, wish I agree is a waste of resources (but can't be easily fixed without extending the Logger api).

blueyed:
> but can't be easily fixed without extending the Logger api

You could set it via private methods directly.

Otherwise your information is what I've seen also.

My plan for an improved caplog fixture would be to e.g. use a raw handler and make that available also.  Haven't looked into it more yet though.

Delgan:
> You could set it via private methods directly.

How could I? It would require to inspect the `format` in order to detect if it contains `"{exception}"`, but I prefer to avoid that kind of parsing.

> My plan for an improved caplog fixture would be to e.g. use a raw handler and make that available also. Haven't looked into it more yet though.

I don't know exactly what you have in mind, improvements are much welcome, as long as they don't extend the api just for uncommon use cases. :wink:

Issue #41, opened 2019-01-20
CI/tests: use tox?


blueyed:
I think it is very useful to have a `tox.ini` to run tests in an isolated way.
This would also allow for using `tox-travis` on Travis then.

What is your opinion?  I'd be happy to provide a PR for this.

Delgan:
I was a little worried that adding `tox` would clutter the repository needlessly and badly integrate with Travis, but thanks to `tox-travis` this is not the case. Using `tox` is best practice anyway, so there is no reason not to use it.

Thanks for the suggestion! A PR would be much appreciated when you have a little time, as you are surely more used to `tox` than I am. 🙂

blueyed:
Cool.
I will also add some flake8 and pytest settings along the way: do you prefer to have them in setup.cfg (or maybe pyproject.toml), or within tox.ini then?

Delgan:
> Cool.
> I will also add some flake8 and pytest settings along the way: do you prefer to have them in setup.cfg (or maybe pyproject.toml), or within tox.ini then?

I wish to limit the number of files and settings to the bare minimum. If it can fit into the `tox.ini`, it's perfect.

How would the integration of `flake8` looks like? This is very useful to improve overall code quality, on the other hand I don't want the CI tests to fail because of styling issues.

blueyed:
Going with tox.ini then.
I would just add some config for flake8, we do not have to inforce it, but it helps already when editing files (I am checking Python files using it by default).  What max-line-length do you prefer?

Delgan:
Yes, `flake8` integration for local development is something definitely missing in my workflow. This will be of great help, thank you!

The max-line-length should be 100.

qcoumes:
@blueyed I can do it if you're to busy.

Delgan:
Closing as it has been fixed. :+1:

PR #94, opened 2019-05-08
Use tox for tests (#41)

qcoumes:
#41 

Both travis and local tests can now be done through `tox`.

Changes :

* Some files modified in `loguru/` to be compliant with PEP 8
* Moving tests dependencies for setup.py to setup.cfg

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 
 dist: xenial
+
 sudo: required
 
 python:
@@ -11,9 +12,10 @@ python:
 
 install:
   - pip install -e .[dev]
+  - pip install tox-travis
+  - pip install codecov
 
-script:
-  - pytest --cov loguru/
+script: tox
 
 after_success:
   - codecov --flags "py${TRAVIS_PYTHON_VERSION//./}"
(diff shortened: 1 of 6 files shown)


qcoumes:
Weird, I also have 94% when I run `pytest --cov` and `coverage report` on *master* on my computer.

Delgan:
Thanks for the work!

The coverage failure is weird, indeed. This may be due to exception formatting tests which are run in a separate process. I guess `pytest` has to use some trickeries to get the coverage right. I would have suggested to replace `py.test` with `pytest --cov` in `setup.cfg`, but you said it failed locally too, so I have no clue. :smile: 

I will give it a try locally this week-end.

Delgan:
I finally take time to reproduce it locally.

I don't know why exactly, but replacing `coverage run` with `pytest --cov`seemed to fix the coverage issue. I think that `coverage run` maybe doesn't handle sub-processes very well.
However, I don't know why you encountered the same problem while running `pytest --cov` on `master`. Perhaps a cache problem or a issue with the python interpreter / path used by `subprocess`? Maybe it will be fixed now that `tox` is used to manage development environment.

Note that I rewrote the history on your branch `qcoumes/tox` to make the corrections. I also added the `E203` error to be ignored by the linter as `black` [automatically adds white-space](https://github.com/python/black#slices) before `:` in slices, which is what is recommended by PEP8 actually.

I may add some changes later (replace `setup.cfg` with `.tox.ini` as I don't know for now if `pyproject.toml` or `setup.cfg` should be preferred, replace `pytest-pep8` with `flake8` or `pylint`, avoid linter to break CI tests), but for now this is very good and will certainly eases my development workflow, thanks @qcoumes!

PR #117, opened 2019-07-22
Fix docstrings

pylipp:
Some minor improvements I came across while reading the docs :)


Delgan:
Great, thank you for taking the time to fix that! :smiley:

PR #136, opened 2019-08-26
Fix a problem of raw option with handler

Writtic:
- Using raw option with handler, the message is cutted last 1 length

When I test my application with loguru, I found something weird.

```
logger.add(
        sink=LogstashHandler(host, port, "my-app", tags),
        format=LOGSTASH_FORMAT,
        level=level
    )
logger.opt(raw=True).error("hello")

Result: {... 'message': 'hell' ...}
```

As you could see, If I use handler with raw option, I couldn't get full message on the log.
So I tried to find this solution then somewhat dealt with it with test code.

### DIFF
--- a/loguru/_handler.py
+++ b/loguru/_handler.py
@@ -108,7 +108,13 @@ class Handler:
 
             if is_raw:
                 if not is_ansi:
-                    formatted = message
+                    if self._is_formatter_dynamic:
+                        formatted = message
+                    else:
+                        if hasattr(self._writer, "read"):
+                            formatted = message
+                        else:
+                            formatted = message + '\n'
                 elif self._colorize:
                     level_ansi = self._levels_ansi_codes[level_id]
                     formatted = self._colorize_format(message, level_ansi)
--- a/loguru/_handler.py
+++ b/loguru/_handler.py
@@ -108,13 +108,7 @@ class Handler:
 
             if is_raw:
                 if not is_ansi:
-                    if self._is_formatter_dynamic:
-                        formatted = message
-                    else:
-                        if hasattr(self._writer, "read"):
-                            formatted = message
-                        else:
-                            formatted = message + '\n'
+                    formatted = message
                 elif self._colorize:
                     level_ansi = self._levels_ansi_codes[level_id]
                     formatted = self._colorize_format(message, level_ansi)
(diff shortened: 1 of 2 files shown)


loguru/_handler.py
```diff
@@ -108,7 +108,13 @@ def emit(self, record, level_id, from_decorator, is_ansi, is_raw):
 
             if is_raw:
                 if not is_ansi:
-                    formatted = message
+                    if self._is_formatter_dynamic:
+                        formatted = message
+                    else:
+                        if hasattr(self._writer, "read"):
```
Writtic: Need to check other conditions

Delgan:
Awesome, thanks for reporting and investigating this issue!

So, to explain what is happening, one should look here: https://github.com/Delgan/loguru/blob/925031720ced203332388696f77a1bd427192ad9/loguru/_logger.py#L716-L717

Why am I chopping the last char of the message? Because Loguru automatically adds `"\n"` at the end of the log messages: https://github.com/Delgan/loguru/blob/925031720ced203332388696f77a1bd427192ad9/loguru/_logger.py#L794

However, standard `logging` does the same: [`Lib/logging/__init__.py#L1084`](https://github.com/python/cpython/blob/6b282e18877ec84e927b381b4ce187eaf4ba3dd7/Lib/logging/__init__.py#L1084)

In order to prevent the message to be displayed with two trailing `"\n"`, I remove the one added by Loguru, but only if the `format` is not a function (because if it is the case, Loguru does not automatically add the trailing `"\n"`).

The bug occurs because the trailing `"\n"` is not added by Loguru if `raw=True`. In such case, there is no `"\n"` to remove, so the last character is wrongly chopped instead.

Here is a small code snippet to play with:

```python
import logging
import sys
from loguru import logger

logger.remove()
logger.add(logging.StreamHandler(sys.stderr), format="{level.no} {message} [Not Chopped]")

logger.error("Allo?")
logger.opt(raw=True).error("hello")
```

Your solution may work but I would like to avoid having to dynamically check for attributes at run-time in `emit()`. Actually, there is a very easy solution: remove `if not is_formatter_dynamic: message = message[:-1]` and replace it with `sink.terminator = ""`. I did not implement this solution is the first time because I wanted not to mutate the sink object, but this is maybe the best solution considering there is no other easy workaround. :confused:

Writtic:
Thank you for very specific and kind replying!
I'm glad to hear the reason of the problem and what your idea about my request is.
I also hoped to avoid amending object level, but I'm not familiar with the project, so I didn't come up with the better solution.

Delgan:
I finally fixed this issue in 49c5560. The workaround was quite simple, I just needed to specialize `"\n"` terminator for both the format and the exception formatter (due to implementation details in `logging.Handler`). It worries me not to have thought about this sooner... :grin: 

Anyway, I merged the new unit tests you added, thanks! :smiley:

PR #165, opened 2019-11-13
make this a one-liner

interrogator:
Seems like this kind of code can be simplified quite a lot...

### DIFF
--- a/loguru/_string_parsers.py
+++ b/loguru/_string_parsers.py
@@ -24,12 +24,16 @@ class Frequencies:
             y, m = t.year + 1, 1
         else:
             y, m = t.year, t.month + 1
-        return t.replace(year=y, month=m, day=1, hour=0, minute=0, second=0, microsecond=0)
+        return t.replace(
+            year=y, month=m, day=1, hour=0, minute=0, second=0, microsecond=0
+        )
 
     @staticmethod
     def yearly(t):
         y = t.year + 1
-        return t.replace(year=y, month=1, day=1, hour=0, minute=0, second=0, microsecond=0)
+        return t.replace(
+            year=y, month=1, day=1, hour=0, minute=0, second=0, microsecond=0
+        )
 
 
 def parse_size(size):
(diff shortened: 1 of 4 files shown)


Delgan:
Yeah, I agree the one-liner looks better, but the "ugly" `if` were actually used on purpose... :slightly_smiling_face: 

I'm not a big fan of using `getattr()` for accessing "static" attributes this way, while the set of valid values is finite and well-known. 

Consider `parse_frequency("__dict__")` and your function will return a `dict` object instead of `None`.

Delgan:
Although I think it would be probably more "pythonic" to build a `dict` with the expected values, rather than chaining `if`:

```python
frequencies = {
    "hourly": Frequencies.hourly,
    ...
}
return frequencies.get(frequency)
```

interrogator:
If you're legitimately worried about something else like `"__dict__"` getting in there, I suppose this PR should ideally fail a test. But yeah, one could also just add:

```python
if frequency not in {'weekly', 'yearly' ...}:
    return
```

I don't mind which style you'd like the most, but surely any of these proposals are nicer and better to maintain than the current branching code...

Delgan:
> If you're legitimately worried about something else like `"__dict__"` getting in there, I suppose this PR should ideally fail a test. But yeah, one could also just add:

Yep, totally, that was my plan after seeing your PR. At the time of writing the tests, I did not not think of this alternative implementation, so I did not add a test to catch this particular edge case (although there is tests for invalid values).



> I don't mind which style you'd like the most, but surely any of these proposals are nicer and better to maintain than the current branching code...

I agree. That's basically a `switch` statement that we need here, and the common idiom for this in Python is to use a `dict`, I think.

Delgan:
Thanks for taking the time to update the PR. :+1: 

I usually run `black` with a line-length of `100`. I guess you probably used the `88` default, which explains why it changed other things.

I gonna be nit-picky but I think I prefer to explicitly construct a `dict` with the expected `Frequencies` values rather than relying on `getattr()`, even at the cost of more lines... It feels "weird" to use the frequencies `set` as a "proxy" to check whether or not `Frequencies` contains the requested value.

I like the refactor of `parse_day()` you made. No need to use regex for such simple function. However, it looks some part of the initial function is missing and broke the CI tests. :grin:

Delgan:
I took the liberty to add a commit to your branch, correcting the few tests and implementing what we had discussed. Thank you for these improvements. :wink:

PR #169, opened 2019-11-20
Expose StrRecord in Public API

t-mart:
I have a custom logging need (sending records to an HTTP API in another thread), so the `CallableSink` seemed pretty attractive to me.

To do my custom logic, I need access to the log record's contextual information, such as level, message, etc, . Thankfully, the argument passed to a `CallableSink` is a `StrRecord` that has a `record` attribute that has all this data as a dict.

However, I can't type hint this argument properly because `StrRecord` is not exposed from the loguru API (it's in a `_`-prefixed package, `loguru._handler.StrRecord`). Therefore, I can't really know if I can access the `record` attribute. So I'm forced to make a leap of faith and hint it as `Any`.

My primary request is that `StrRecord` is exposed from the loguru API so I can safely access the `record` attribute. I think this is reasonable, if not necessary, considering that they're passed to user code and is also [a documented feature](https://loguru.readthedocs.io/en/stable/api/logger.html#message).

- [x] Expose `StrRecord` in public API
- [x] Create `StrRecord` docstring
- [x] Create documentation page for `StrRecord`
- [x] Reference `StrRecord` in docs where appropriate
- [x] Add change to `CHANGELOG.rst`

**I definitely need some review on my docstring and Sphinx formatting and style.**

### Versions
loguru version `0.3.2`
Python version `3.8.0`
Windows 10, 64-bit

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@
 - Add the level ``name`` as the first argument of namedtuple returned by the ``.level()`` method.
 - Replace ``ValueError`` with ``TypeError`` for exceptions raised when a ``logger`` method was called with argument of invalid type.
 - Remove inheritance of some record dict attributes to ``str`` (for ``"level"``, ``"file"``, ``"thread"`` and ``"process"``).
+- Expose ``StrRecord`` in the public API so that custom sink functions may be properly typed.
 
 
 `0.3.2`_ (2019-07-21)
--- a/docs/api/strrecord.rst
+++ b/docs/api/strrecord.rst
@@ -1,0 +1,7 @@
+loguru.StrRecord
+================
+
+.. automodule:: loguru._handler
+
+.. autoclass:: loguru._handler.StrRecord()
+   :members:
--- a/loguru/__init__.py
+++ b/loguru/__init__.py
@@ -8,6 +8,7 @@ import sys as _sys
 
 from . import _defaults
 from ._logger import Logger as _Logger, Core as _Core
+from ._handler import StrRecord
 
 __version__ = "0.3.2"
 
--- a/loguru/_handler.py
+++ b/loguru/_handler.py
@@ -10,6 +10,15 @@ from ._ansimarkup import AnsiMarkup
 
 
 class StrRecord(str):
+    """
+    A record to be logged.
+
+    This type is a subclass of |str| and may be used as such in cased where the formatted
+    message is desired.
+
+    Instances of ``StrRecord`` have an attribute ``record`` which is a |dict| containing all 
+    contextual information possibly needed.
+    """
     __slots__ = ("record",)
 
 
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -329,8 +329,9 @@ class Logger:
           method, it will be automatically called at sink termination.
         - A file path as |str| or |Path|. It can be parametrized with some additional parameters,
           see below.
-        - A simple |function|_ like ``lambda msg: print(msg)``. This allows for logging
-          procedure entirely defined by user preferences and needs.
+        - A simple |function|_ like ``lambda msg: print(msg)``. Will be called with one positional
+          argument of type |StrRecord|. This allows for logging procedure entirely defined by user
+          preferences and needs.
         - A built-in |Handler| like ``logging.StreamHandler``. In such a case, the `Loguru` records
           are automatically converted to the structure expected by the |logging| module.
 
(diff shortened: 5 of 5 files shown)


Delgan:
Hey @t-mart, thanks for contributing to Loguru! 

Being able to type hint sinks functions is very important. However, I'm not sure I want to expose `StrRecord` in the documented API. I we move in that direction, that would also require to add trivial classes like [record attributes](https://github.com/Delgan/loguru/blob/4408a33f2c0de31f93777e455a9a04caf070e1c9/loguru/_recattrs.py) and the [`Catcher`](https://github.com/Delgan/loguru/blob/4408a33f2c0de31f93777e455a9a04caf070e1c9/loguru/_logger.py#L972) context-decorator to the public API.

I would like to avoid cluttering the documentation with all these insignificant objects. Main goal was to keep the API minimal. Also, I don't see much interest in `from loguru import StrRecord` (besides type hints), that's why I prefer not to make it directly accessible.

Still, as you said, it's useful for type hinting. Do you think it's reasonable to make `StrRecord` and its peers only exposed to the "type hints API"?

I recently added them in the form of a stub file [`__init__.pyi`](https://github.com/Delgan/loguru/blob/4408a33f2c0de31f93777e455a9a04caf070e1c9/loguru/__init__.pyi). From what I tested, it seems to work fine:

```python
from __future__ import annotations  # Python < 3.8

import loguru

def good_sink(message: loguru.Message):
    print(message.record)

def bad_sink(message: loguru.Message):
    print(message.invalid)

loguru.logger.add(good_sink)
loguru.logger.add(bad_sink)
```
```bash
$ mypy test.py
test.py:9: error: "Message" has no attribute "invalid"
Found 1 error in 1 file (checked 1 source file)
```

Actually, it's much better to use such annotations rather than importing the type, because it also permits to use the [`Record`](https://github.com/Delgan/loguru/blob/4408a33f2c0de31f93777e455a9a04caf070e1c9/loguru/__init__.pyi#L72) type and hence annotate a `filter` function for example (and even `StrRecord.record`). That would not be possible otherwise, as there is no `Record` type to import.

However, you made me realize two things:
* There are inconsistencies between class and type names (`StrRecord == Message`)
* The type hints are nowhere documented nor referenced, so there is no indication that they even exist

I will make sure to fix the naming issues. I'm not sure how to properly document them, though. I considered adding a link to `__init__.pyi` directly at the top of the API Reference, and arranging the file so that it is more self-documenting (by adding some comments for example). There is several useful custom types in this stub file, but I don't imagine adding documentation for all of them, so linking directly to the file seems to be an acceptable compromise.

t-mart:
@Delgan 

- > Do you think it's reasonable to make StrRecord and its peers only exposed to the "type hints API"?

    Yes! I didn't know that was a thing! That works for me. And I agree about the avalanche effect exposing StrRecord might have for other types.

    Actually, I didn't even know `Message` existed. When I `type()`-ed the argument to my sink function, it said it was an `StrRecord`.

    ```python
    from __future__ import annotations  # Python < 3.8

    import loguru

    def good_sink(message: loguru.Message):
        print(type(message))

    loguru.logger.remove()
    loguru.logger.add(good_sink)

    loguru.logger.info("My message")

    # prints the following:
    #
    # <class 'loguru._handler.StrRecord'>
    ```

  Are you sure `Message` is right here? I'm not sure if you're proposing to expose `StrRecord` as a type hint named `Message`, or if `Message` is part of the current API that I missed. 

- > I'm not sure how to properly document them, though.

  Haha, well, nor do I, but I'd trust your efforts!

Delgan:
Yeah, it's a recent feature, I have not yet published a new version including them. As documentation about type hints is missing, I'm not surprised you wasn't aware of this. :smile:

>  Are you sure `Message` is right here? I'm not sure if you're proposing to expose `StrRecord` as a type hint named `Message`, or if `Message` is part of the current API that I missed.

I was thinking to rename internal `StrRecord`  to `Message` (so that it matches the type hint name). It looks more explicit and elegant. Currently, the message object uses a [`StrRecord`](https://github.com/Delgan/loguru/blob/4408a33f2c0de31f93777e455a9a04caf070e1c9/loguru/_handler.py#L12) class, but it's annotated as a [`Message`](https://github.com/Delgan/loguru/blob/4408a33f2c0de31f93777e455a9a04caf070e1c9/loguru/__init__.pyi#L87) class in the stub file.

>    Haha, well, nor do I, but I'd trust your efforts!

I will add several comments to the `__init__.pyi` file, I add links to this file at the top of the API reference and in the Readme, and I will add an example among the other [code snippets](https://loguru.readthedocs.io/en/stable/resources/recipes.html) in docs. Thanks for bringing this issue to my attention. :+1:

Delgan:
For reference, types have been added to the documentation here: [**Type hints**](https://loguru.readthedocs.io/en/latest/api/type_hints.html). :wink:

PR #174, opened 2019-11-23
Give queue worker thread a name.

t-mart:
Hey @Delgan,

I'm doing some debugging with a multi-threaded application and also have a Loguru sink with `enqueue=True`. (As we've discussed before) this instructs Loguru to manage a thread that processes a queue of log records.

Currently, that thread's name is the default one:
> By default, a unique name is constructed of the form “Thread-N” where N is a small decimal number.

_[https://docs.python.org/3.8/library/threading.html#threading.Thread](https://docs.python.org/3.8/library/threading.html#threading.Thread)_

For identification purposes, it'd be helpful to see a descriptive name for this thread in environments/debuggers/etc that support listing threads.

The name I've chosen `loguru-writer`.

This doesn't affect the operation of the Loguru in any other way:

> [The name is] A string used for identification purposes only. It has no semantics.

_[https://docs.python.org/3.8/library/threading.html#threading.Thread.name](https://docs.python.org/3.8/library/threading.html#threading.Thread.name)_

- [x] Give queue worker thread a name.
- [x] Add change to CHANGELOG.rst.
- [x] Tox passes.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,7 +15,7 @@
 - Add the level ``name`` as the first argument of namedtuple returned by the ``.level()`` method.
 - Replace ``ValueError`` with ``TypeError`` for exceptions raised when a ``logger`` method was called with argument of invalid type.
 - Remove inheritance of some record dict attributes to ``str`` (for ``"level"``, ``"file"``, ``"thread"`` and ``"process"``).
-
+- Give queue worker thread a name.
 
 `0.3.2`_ (2019-07-21)
 =====================
--- a/loguru/_handler.py
+++ b/loguru/_handler.py
@@ -68,7 +68,7 @@ class Handler:
             self._thread = threading.Thread(
                 target=self._queued_writer,
                 daemon=True,
-                name="loguru-writer"
+                name="loguru-writer-%d" % self._id
             )
             self._thread.start()
 
--- a/loguru/_handler.py
+++ b/loguru/_handler.py
@@ -65,7 +65,11 @@ class Handler:
         if self._enqueue:
             self._owner_process = multiprocessing.current_process()
             self._queue = multiprocessing.SimpleQueue()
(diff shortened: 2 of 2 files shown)


t-mart:
Hmm. It seems I've caused code coverage to drop. Would you help me interpret why?

[The drop](https://codecov.io/gh/Delgan/loguru/compare/44fe9343524cd9eaaeaff2f6689c57161f46a507...a59b1810c6cafa2558e85220e4959debf3bce5fb/changes#L102) seems to stem from a blank line?

Delgan:
Good idea, thanks for the contribution!

What do you think of appending the handler `_id` to the thread name, like `name="loguru-writer-%d" % self._id`? This would give even more information, and avoid confusion if there are several running threads due to multiple sinks added with `enqueue=true`.

I don't know while coverage check failed... I encountered this problem once, on this same line. This line is executed in only one Travis build (Python 3.5.2), maybe there is a conflict with another build (Python 3.5). Particularly, the [`.travis.yml`](https://github.com/Delgan/loguru/blob/44fe9343524cd9eaaeaff2f6689c57161f46a507/.travis.yml) lists six builds while there is only five in the [`tox.ini`](https://github.com/Delgan/loguru/blob/44fe9343524cd9eaaeaff2f6689c57161f46a507/tox.ini) file. Anyway, I restarted the build and it worked fine. :slightly_smiling_face:

t-mart:
> What do you think of appending the handler _id
Sounds good, as long as you are sure that will always be a successful interpolation (i.e. handlers always have an `_id` attribute?).

Delgan:
Yeah, don't worry, I will update the PR if you don't have the time. :+1:

t-mart:
@Delgan, the `-<id>` suffix has been added as you suggested.

Delgan:
Great, thanks a lot!

PR #176, opened 2019-11-23
Handle quoted environment variables from Windows 10

yoonghm:
To set an environment with < in Windows 10 must be quoted with ", for example 
set LOGURU_CRITICAL_COLOR="<RED><bold>"

However, os.environ("LOGURU_CRITICAL_COLOR") would return '"<RED><bold>"'.  The changes replace the outermost single quotes.

### DIFF
--- a/loguru/_defaults.py
+++ b/loguru/_defaults.py
@@ -1,3 +1,4 @@
+import os
 from os import environ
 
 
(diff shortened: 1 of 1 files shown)


Delgan:
Thanks for the pull request. I was not aware that `"<"` has a special meaning while using with `set`. However, I discovered there is a way to escape such symbol by prepending it with `"^"`.

```bash
set LOGURU_CRITICAL_COLOR ^<green^>
```
Then it works fine:
```python
>>> import os
>>> os.environ["LOGURU_CRITICAL_COLOR"]
'<green>'
```
Therefore I don't think the Loguru `env()` function should be changed, it's something that should be handled in the OS environment.

Issue #192, opened 2019-12-23
Can't chain .opt() calls


gazpachoking:
Calling opt more than once on a logger instance overrides all options that aren't specified to default, meaning you can't chain .opt() calls together.
For example in the following call, the color codes are ignored.
```python
logger.opt(ansi=True).opt(depth=1).debug('<red>aoeu</>')
```
This doesn't make a lot of sense when done directly like this, but if you are using the [recipe for adding custom loglevel methods](https://github.com/Delgan/loguru/blob/master/docs/resources/recipes.rst#using-logging-function-based-on-custom-added-levels), it calls .opt() to set the depth. The consequence of this is that you can no longer use .opt() directly from you code, since it will always be overridden by the depth call, including all of the options other than depth.

Delgan:
Huh, this is a serious design issue indeed. Thanks for the report and the pull request.

I'm not a big fan of sentinel values, as it makes the signature of the function less readable. I wonder if there is any other possible workaround... If not, I will probably merge your fix. :+1:

gazpachoking:
Could switch to `**kwargs` instead of having a sentinel value... but that makes the signature even worse. Not sure a better way. At least the signature in the pyi file is still nice with type hints.

Delgan:
I don't think there's much point in chaining `opt()` calls, apart from the problem with custom levels you've raised. So, what do you think of replacing the documented recipe with:

```python
from functools import partialmethod

logger.__class__.foobar = partialmethod(logger.__class__.log, "foobar")
```

I was delightfully surprised to find out that it worked just fine! I guess the Python interpreter somehow optimize [`partialmethod()`](https://docs.python.org/3/library/functools.html#functools.partialmethod) so it does not add another frame to the stack, thus avoiding the need for `opt(depth=1)`. Then the `foobar()` logging method can be combined with any `opt()` attribute.

gazpachoking:
I think that recipe change sounds like a good plan regardless of whether opt gets changed.
I sorta like the idea of it working as expected when chained, but it's much less important with that updated recipe. The only other instance I can think of is if someone perhaps wanted to do something like turn on colors by default when logging in only one of their modules. e.g.

```python
from loguru import logger

logger = logger.opt(ansi=True)


try:
    somethingdangerous()
except:
    logger.opt(exception=True).debug('Everything <blue>went</> wrong')
```

Delgan:
It seems possible to achieve the desired behavior using the same principle:

```python
from functools import partial

[...]

PR #193, opened 2019-12-23
Allow .opt() calls to be chained on loggers

gazpachoking:
This addresses issue #192 
Calling .opt() multiple times now keeps the options set on previous calls rather than setting all unspecified options back to their default.

Wasn't sure best how to make the test for this, so I just copied another one that used multiple options, and split it apart into 2 `.opt()` calls.


Delgan:
Closing due to discussions in #192.

The documentation has been updated with a new workaround for custom levels.

Issue #195, opened 2019-12-30
Log deletion too greedy


gazpachoking:
Using the file sink, when loguru goes to clean up old logs due to retention setting. It is a bit too greedy with what it thinks are old logs. I am using the filename `flexget.log`, and the glob for deletion comes out to `flexget.*`. This strikes me as a bit greedy, and in my particular case it tries to delete a folder called `FlexGet.egg-info` (which is read only and causes a crash.) The two easy improvements I can see are:
- Include the extension in the glob
- Exclude folders

Delgan:
Good catch, thanks. The filenames can take many forms due to `"{time}"`, `rotation` and `compression`. I will modify the collect of files based on your two suggestions. In the meantime, I guess you will need to manually re-filter and remove files using a custom `retention` function. :wink:

Delgan:
Closed by #196.

PR #196, opened 2019-12-30
Retention fix

gazpachoking:
Addresses issues in #195 

- Ensures directories are never deleted when handling retention
- Ensures log extension is included in filenames when handling retention

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,9 @@
+`Unreleased`
+============
+
+- Prevent unrelated files and directories to be incorrectly collected thus causing errors during the `retention` process (`#195 <https://github.com/Delgan/loguru/pull/195>`_, thanks `@gazpachoking <https://github.com/gazpachoking>`_).
+
+
 `0.4.0`_ (2019-12-02)
 =====================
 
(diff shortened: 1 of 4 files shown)


gazpachoking:
Oops, should have run the tests locally before submitting. Requiring the dots was the one thing I knew could be made more specific, and is what's causing the failure right now. I was trying to avoid changing too much, I think to make it better we need multiple glob patterns.

Delgan:
Thanks for the PR!
I guess it's just a small edge case, I fix is probably possible while keeping just one glob pattern. I will take a look.

Delgan:
@gazpachoking Ok, you was totally right, multiple glob patterns are required. 

I took the liberty of making changes to your branch as a result of this PR, I hope you don't mind. 

While refactoring the code, I decided to combine the glob pattern with a regex check which I think is more permissive: we want to match `"file.2020-01-01_01-01-01_0000001.log"` (automatic renaming on rotation), but we want to exclude both `"file.foo.log"` and `"foo.file.log"` which is complicated without regex.

At the same time, I realized it was wrong to call `os.path.splitext()` after  `glob.escape()` as it would result in something like `("file\\", ".log")`, so I modified the corresponding code. 

Also, I tried to fix a 100%-theoretical bug that could occur if `"."` was present in the `{time}` spec format (leading to wrong `splitext()`).

This results in code much more convoluted unfortunately, but this should greatly reduce false positives during `retention` process. Does it look good to you?

gazpachoking:
It is a bit hard to read and understand the glob and regex bit, but the expanded test suite is clear, and makes me confident it's doing what it's supposed to now. 👍

Delgan:
Great, thanks for the review! Let's merge it then. ;)

PR #211, opened 2020-02-07
fix gunicorn --workers=4 rotation create 4 log file

rxy007:
test.py:
```
from flask import Flask
from flask.json import jsonify
import os
from loguru import logger
app = Flask(__name__)
pid = os.getpid()
logger.remove()
logger.add('logs/log', format='{time} {level} {module} {function} {line} {message}', catch=True, type='json', retention='30 days', rotation='12:00', encoding='utf8', enqueue=True)
@app.route('/log', methods=['GET'])
def parse():
    logger.info(pid)
    return jsonify({'pid': pid})
```

if i run test.py ：
` gunicorn --workers=4 --threads=4 --worker-class=gthread -b :5000 test:app`

rotation after:
create four log file:
image


Delgan:
Hi @rxy007.

I don't know much about `gunicorn`, but I guess that 4 log files are created because `logger.add()` is placed is not guarded by any `if __name__ == "__main__"` condition and instead placed at the "root" level of your module. I suppose `gunicorn` needs to start a new process for each worker, which means it will need to reload your entire module, leading to `logger.add()` being called four times.

You probably need to rearrange your code according to `gunicorn` standards so that `logger.add()` is called only once, no matter how many processes are started.

rxy007:
Hi Delgan:
you are right, gunicorn needs to start a new process for each worker and I can't control logger.add() is called only once,  so i changed the method of rotation, For example, the rotation is timing. When the rotation is satisfied before, I will check the latest modification time of the log file. If the latest modification time is after the rotation, it means that the file has been created by another process, and there is no need created



xiaoyu.r@aliyun.com
 
From: Delgan
Date: 2020-02-08 02:48
To: Delgan/loguru
CC: 冉小瑜; Mention
Subject: Re: [Delgan/loguru] fix gunicorn --workers=4 rotation create 4 log file (#211)
Hi @rxy007.
I don't know much about gunicorn, but I guess that 4 log files are created because logger.add() is placed is not guarded by any if __name__ == "__main__" condition and instead placed at the "root" level of your module. I suppose gunicorn needs to start a new process for each worker, which means it will need to reload your entire module, leading to logger.add() being called four times.
You probably need to rearrange your code according to gunicorn standards so that logger.add() is called only once, no matter how many processes are started.

Delgan:
@rxy007 This is an acceptable fix for your use case, but I don't think it generalizes well. The correct way to fix that is to ensure `logger.add()` is called only once. How to do so? I don't know, it probably depends of some `gunicorn` setup and code rearrangement. Maybe ideas for a solution can also be found in [guide to use `multiprocessing` with `loguru`](https://loguru.readthedocs.io/en/stable/resources/recipes.html#compatibility-with-multiprocessing-using-enqueue-argument).

Issue #271, opened 2020-05-21
KeyError: "Attempt to overwrite 'args' in LogRecord"  


kornicameister:
### Expected result:

I am able to use `args` as keyword to denote argument to string interpolation when using loguru

### Actual result:

Opposite of above ;-).In fact it seems that if you try to specify whatever is originally available in Record from `logging` of python, python will think you are crazy to put such things in record.

### Facts

- `loguur.__version__ = 0.5.0`
- `python --version = Python 3.7.6`
- `lsb_release -sd = Linux Mint 19.3 Tricia`
- `context = application`

### Reproduce

```
git clone https://github.com/kornicameister/axion.git /tmp/axion --depth 1 ; \
cd /tmp/axion ; \
git checkout c9d85e76bd48719aab7edded1d4287a52ecb145e \
command -v tox && tox -e "py37-test" -- -k test_not_empty_signature || echo "need tox" ; \
cd - \
rm -rf /tmp/axion
```

### Root cause

Relevant line that causes issue available [here](https://github.com/kornicameister/axion/blob/60659e81ba4666a966d119ed7069da13e8802a0b/axion/handler/analysis/__init__.py#L56)

Alternatively, it might be caused by `conftest#caplog` [here](https://github.com/kornicameister/axion/blob/60659e81ba4666a966d119ed7069da13e8802a0b/tests/conftest.py#L30), but this is unlikely given that only one test fails ;-)

Delgan:
Hi @kornicameister, thanks for the detailed bug report!

Since version `0.5.0` of Loguru, keyword arguments when calling a logging function are automatically added to the `extra` dict. This feature has been requested multiple times to facilitate structured logging.

I didn't think it could have that side effect. This happens because if you use a handler from the standard `logging` library, [the `extra` dict is used to update the `LogRecord`](https://github.com/python/cpython/blob/b831129123dbb4bfba49824ae893448b05398f27/Lib/logging/__init__.py#L1546-L1550). This happens also if `logger.bind(name="name")` is used for example.

Unfortunately there's not much I can do. While using a standard `Handler`, one have to be careful not to use "forbidden" keywords. Possible workarounds:
* Rename `args` to `args_`
* Use f-strings to format the message
* Specify `logger.opt(capture=False)`

Also, I noticed in your code base you're using `opt(record=True)` but I'm not sure you need it if you don't use `logger.info("Current thread: {record[thread]}")` for example.

kornicameister:
[...]

PR #276, opened 2020-05-27
Implement mypy plugin for loguru

kornicameister:
Closes: #271


kornicameister:
@Delgan I took some liberty in designing how messages will look like. We can always tweak or adjust that. Waiting for tour feedback.

tox.ini
```diff
@@ -11,7 +11,11 @@ commands =
     coverage report -m
 
 [pytest]
-addopts = -l
+addopts = -l --mypy-same-process --mypy-ini-file=typesafety/mypy.ini
```
kornicameister, line 14: that's quite awkward...`tox` for some reason chokes on that :/. It shouldn't, I know for sure it works because this is what I have been using for some time.

tox.ini
```diff
@@ -11,7 +11,11 @@ commands =
...
+testpaths =
+    tests/
+    typesafety/
+log_level = DEBUG
```
kornicameister, line 18: We can drop that log_level I believe. Sorry for not doing the before.

kornicameister:
@Delgan do you have any idea how to fix that travis situation? I am not much of a travis user myself TBH :(

Delgan:
Hey @kornicameister, thank you so much for all the work you've done!

I had absolutely no idea it was possible to achieve such a level of accuracy with `mypy` plugins. I tested your plugin and it totally blows my mind how you are able to detect bad usage of arguments. That's really great work.

I don't know what's wrong with Travis but I actually encounter the same error on my local computer. While running `tox`, `--mypy-same-process` and `--mypy-ini-file` arguments are not recognized  although `pytest-mypy` and `pytest-mypy-plugins` are both installed. The `pytest` command works fine without `tox`, though. :confused:

So, I didn't know what a mypy plugin looked like back to a few days. I'm quite impressed you managed to make something fully working without even needing to modify the  `loguru` source code. 

However, I don't think I want to maintain all of this with as much dedication as I'm trying for the rest of the codebase. Please tell me: what is your opinion about creating a new repository and library for `loguru-mypy`?

Personally, I would feel more comfortable that way. Managing such a plugin involves a lot of new things to look after, there are potential bugs [1] and edge cases [2]. Not being myself a user of `mypy`, I am reluctant to take the responsibility of related developments.

Of course the plugin will be mentioned in the documentation to ease discovery to those who may need it. What do you think?

-----

[1] _This raises an exception inside the plugin:_
```python
from operator import add
from loguru import logger

logger.info("Crash: {}", add)
```


[2] _This is reported as an error while it shouldn't:_
```python
from loguru import logger

def foo(a, b):
    a / b

try:
    foo(1, 0)
except Exception:
    logger.exception("An error occurred while executing: '{}'", foo)
```

kornicameister:
Plugins are normally not enabled by default, so this could mean we can have external repository. On the other hand that makes plugin harder to discover. Still...I have seen plugins done both ways. `pydantic` does embedding , but you find external repositories for SQLAlchemy though.

This is your repo, but if this plugin allows to avoid runtime problems, I would dare to suggest embedding and maintaing here. Otherwise this can go in external repo and be linked here as another `extra`.

Personalny, I would prefer external repo just to get które "fame". But this is quite selfish and I am not realny sure if that is better. TBH I wonder if checks match internal logic of loguru, which would be another reason to maintain plugin close to sources.

Delgan:
I don't think there's a big discoverability problem because we have to mention at some point that a `mypy` plugin exists anyway, it's not automatic.

I agree that ideally it would be better to have all functionalities in one place. That would probably be the best. However, the more I think about it, the more I'd prefer it to be in a separate repository. At first I was not sure, but it's too much work otherwise. I'd be happy to contribute if I have the time though, but I'd have a "freer" mind.

As I prefer not to have issues reported directly here, I think it should be explicitly installed separately instead of using `extra`. And add a new paragraph in [the documentation about type hints](https://loguru.readthedocs.io/en/stable/api/type_hints.html) to mention the library and its usage.

Sorry I don't feel like integrating it directly into the codebase, but it's easier for me and it won't stop people from benefiting from your work.

kornicameister:
No worries. I wilk create a repo under my name. I was thinking about loguru organization for a while. But..yeah, we can always do that anytime later.

kornicameister:
@Delgan let's move entire discussion to linked PR. This PR; I am closing this PR now, 
Later on, I will create a new one to advertise new plugin in here.

Delgan:
Yep, great! Thanks again. :+1: 

I opened an new issue about documenting the `loguru-mypy` plugin: https://github.com/Delgan/loguru/issues/277

PR #287, opened 2020-06-15
Add structured data to caplog

benhowes:
Suggesting this because I thought it would be useful for people who start using structured params - I've checked that it works for me.

Thanks :)


Delgan:
I don't really use that feature, so I'm going to trust you. Thank you! :grin:

PR #323, opened 2020-09-13
docs: Fix simple typo, propogates -> propagates

timgates42:
There is a small typo in docs/resources/migration.rst.

Should read `propagates` rather than `propogates`.


Delgan:
Thank you for taking the time to open a PR and correct this! :+1:

Issue #309, opened 2020-08-19
Loguru doesn't remove uncompressed files after compressing them


dstlmrk:
Hello,

I use this definition for compressing but I expected different behaviour:

```python
handlers.append({
    "rotation": "03:00",
    "compression": "gz",
    "retention": 10 days",
    ...
})
```

All my raw log files are compressed every night. It's working. But I don't need them (uncompressed files) if they are already compressed. Loguru doesn't remove them and I don't know if it's wanted behaviour. Because now it's not saving any space on disc.

Should I use my custom callable which removes all old uncompromised files?

Delgan:
Hum... That's not the expected behavior. The `.log` file should be removed once compression is done. May I ask you which `loguru` version you're using please? You don't see any error message in your logs (I'm thinking to permission issues for example)? You may try to create a minimal script with `rotation="10 kB"` and a basic `for` loop to see if you can reproduce the problem consistently and hence investigate it. :confused:

dstlmrk:
Hi! I will close this issue in next days after test probably. I was getting this traceback in uwsgi log:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/loguru/_handler.py", line 176, in emit
    self._sink.write(str_record)
  File "/usr/local/lib/python3.8/site-packages/loguru/_file_sink.py", line 174, in write
    self._terminate_file(is_rotating=True)
  File "/usr/local/lib/python3.8/site-packages/loguru/_file_sink.py", line 210, in _terminate_file
    self._compression_function(old_path)
  File "/usr/local/lib/python3.8/site-packages/loguru/_file_sink.py", line 66, in compression
    compress_function(path_in, path_out)
  File "/usr/local/lib/python3.8/site-packages/loguru/_file_sink.py", line 53, in copy_compress
    with open(path_in, "rb") as f_in:
FileNotFoundError: [Errno 2] No such file or directory: '/var/log/app/app_2020-09-07_12-34-24_908563.log'
```

Then I found the issue #220 and there is info about [enqueue](https://github.com/Delgan/loguru#asynchronous-thread-safe-multiprocess-safe) param. Because my app is multiprocessing it makes sense I need the `enqueue=True` :)

dstlmrk:
Ok that is bigger problem than I thought :( If I want have only one logging file for whole app instead of file for each process then I need `enqueue=True`. But there is problem with killing processes. For this example I created simple web server with Flask and uwsgi:

```python
# main.py

[...]

PR #325, opened 2020-09-14
Fix removing simple queue in child processes

dstlmrk:
Hello, I think this pull request could fix https://github.com/Delgan/loguru/issues/309.

There is problem with calling `multiprocessing.current_process()` which returns the instance of `_MainProcess()`. But because this code is executed before a new thread there are same values/objects. In child process is the same object as in parent process. You can check it with `id(self._owner_process)`.

Source code of multiprocessing: https://github.com/python/cpython/blob/master/Lib/multiprocessing/process.py#L41

Another way how to fix it is to move `self._owner_process = multiprocessing.current_process()` after creating a new thread. There would be different objects in every process. But I think there is better to work wih PID numbers than objects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,6 @@ ENV/
 
 # VScode
 .vscode/
+
+# Idea IDE
+.idea/
--- a/loguru/_handler.py
+++ b/loguru/_handler.py
@@ -1,5 +1,6 @@
 import functools
 import json
+import os
 import multiprocessing
 from threading import Thread
 
@@ -66,7 +67,7 @@ class Handler:
         self._queue = None
         self._confirmation_event = None
         self._confirmation_lock = None
-        self._owner_process = None
+        self._owner_process_pid = None
         self._thread = None
 
         if self._is_formatter_dynamic:
@@ -85,7 +86,7 @@ class Handler:
             self._queue = multiprocessing.SimpleQueue()
             self._confirmation_event = multiprocessing.Event()
             self._confirmation_lock = multiprocessing.Lock()
-            self._owner_process = multiprocessing.current_process()
+            self._owner_process_pid = os.getpid()
             self._thread = Thread(
                 target=self._queued_writer, daemon=True, name="loguru-writer-%d" % self._id
             )
@@ -184,7 +185,7 @@ class Handler:
         with self._lock:
             self._stopped = True
             if self._enqueue:
-                if self._owner_process != multiprocessing.current_process():
+                if self._owner_process_pid != os.getpid():
                     return
                 self._queue.put(None)
                 self._thread.join()
@@ -201,7 +202,7 @@ class Handler:
             self._confirmation_event.clear()
 
     async def complete_async(self):
-        if self._enqueue and self._owner_process != multiprocessing.current_process():
+        if self._enqueue and self._owner_process_pid != os.getpid():
             return
 
         with self._lock:


Delgan:
Wow! Thanks a lot for taking the time to investigate and fix this issue!

Kudos for finding the solution. To be honest I'm still scratching my head trying to figure out what's going on. :astonished: 

> In child process is the same object as in parent process.

Why that?
Why does calling `multiprocessing.current_process()` in the child returns the same process as in the parent?
Why does it fail if `_owner_process` is a `Process()` object but not if it's an `init`? Is `uwsgi` somehow mutating the main process?
Why does it freeze at `stop()` anyway? 
 
It will take me probably some time to properly grasp all of that, but I will publish your correction as soon as possible. Thank you very much!

dstlmrk:
I will try to describe it :) And I want to apologize because last part "Another way how to fix it..." of my previous comment is not correct.

#### 1), 2) Why does calling multiprocessing.current_process() in the child returns the same process as in the parent? and 3b) Why does it fail if _owner_process is a Process() object but not if it's an int?

Why multiprocessing returns the same object? Because of [this](https://github.com/python/cpython/blob/master/Lib/multiprocessing/process.py#L41) and [this](https://github.com/python/cpython/blob/master/Lib/multiprocessing/process.py#L416). This object/instance is created once in master process and everytime when you call `multiprocessing.current_process()` you get the same instance (it doesn't create new instance everytime when you call it). That's property of multiprocessing. It returns the same object but it behaves differently in every process. You can call [property](https://github.com/python/cpython/blob/a33f2c2bae759fc9d06e1c032fd2026135f2df45/Lib/multiprocessing/process.py#L235)  `pid` (`ident`) and it will return pid of current process. That's the reason why `self._owner_process` doesn't return pid of parent process. Actually `self._owner_process` = `multiprocessing.current_process()` everytime.

I can show you my custom output from code (everywhere is the same object but it behaves differently):

```
...
[os.getpid()=19870] I'm MASTER!!!
[os.getpid()=19870] In Handler.__init__() id(self._owner_process)=139955875152368
[os.getpid()=19870] In Handler.__init__() self._owner_process.pid=19870
WSGI app 0 (mountpoint='') ready in 0 seconds on interpreter 0x55630ead8a50 pid: 19870 (default app)
*** uWSGI is running in multiple interpreter mode ***
spawned uWSGI master process (pid: 19870)
spawned uWSGI worker 1 (pid: 19873, cores: 1)
spawned uWSGI http 1 (pid: 19874)
^CSIGINT/SIGQUIT received...killing workers...
gateway "uWSGI http 1" has been buried (pid: 19874)
[os.getpid()=19873] Exit process
[os.getpid()=19873] In Handler.stop() id(self._owner_process)=139955875152368 ?= id(multiprocessing.current_process())=139955875152368
[os.getpid()=19873] In Handler.stop() self._owner_process.pid=19873 ?= multiprocessing.current_process().pid=19873
Thu Sep 17 12:51:19 2020 - worker 1 (pid: 19873) is taking too much time to die...NO MERCY !!!
...
```

#### 3a)  Is uwsgi somehow mutating the main process?

The `uwsgi` doesn't mutate the main process.

#### 4) Why does it freeze at stop() anyway?

[This condition](https://github.com/Delgan/loguru/blob/master/loguru/_handler.py#L187) means children should return and don't continue in code. But every child thinks he is master so every child kills logging thread (simple queue) and I don't know exactly why but it causes timeout in children/worker proccesses.

Delgan:
Thanks for the explanations!

One thing is still confusing me.

> This object/instance is created once in master process and everytime when you call `multiprocessing.current_process()` you get the same instance (it doesn't create new instance everytime when you call it). 
> [...]
> Actually `self._owner_process = multiprocessing.current_process()` everytime.

That's not what I'm observing while using standard `multiprocessing` without involving `uwsgi`:

```python
import multiprocessing

main = multiprocessing.current_process()

def hello_world():
    child = multiprocessing.current_process()
    print("main == child:", main == child)

if __name__ == "__main__":
    p = multiprocessing.Process(target=hello_world)
    p.start()
    p.join()
```

This prints `"main == child: False"`. 

Using `uwsgi` I get the opposite result: 

```python
import multiprocessing
from flask import Flask

application = Flask(__name__)
main = multiprocessing.current_process()

@application.route('/')
def hello_world():
    child  = multiprocessing.current_process()
    return "main == child: {}".format(main == child)
```

That's this behavior difference that I can't understand. :confused:

dstlmrk:
Hmm.. I checked multiprocessing code and I see reason probably there. This library does some overhead before starting process. I see it calls `_bootstrap` method, [replaces](https://github.com/python/cpython/blob/master/Lib/multiprocessing/process.py#L301) `_current_process` with `self` (the child process) and [deletes](https://github.com/python/cpython/blob/master/Lib/multiprocessing/process.py#L312) the original `current_process`. I think that's the reason why child process in multiprocessing has new instance every start of `Process`.

I'm not sure but `uwsgi` doesn't use multiprocessing library probably and can't execute this overhead himself. In our applications based on uwsgi we can't use loguru without this change.

Delgan:
Yeah, `uwsgi` probably implements its own bootstrapping procedure. :confused:

Anyway, thanks again. I wanted to be sure it was somehow related to `uwsgi` and  that `loguru` was working properly out of that context.

I will merge this PR and probably publish a new version during the week-end. :+1:

dstlmrk:
I am glad I could help you :)

Delgan:
Again, thanks a lot for your help! It should be fine in `v0.5.3`. :+1:

Issue #353, opened 2020-10-21
Code cleanup and fix flake8 errors


AndrewYakimets:
Hi, I am having a big desire to make my first contribution. And I would really like to start with helping loguru! I saw there is some unused imports (for e.g., setup.py:1), code that is hard to read and can be improved (loguru/__init__.py:17 - add key/value to params) and some amout of flake8 errors (mostly in tests). I would like to fix it and make the contribution to loguru.

Could you, please, tell me if I am allowed to do so and you don't mind?

Delgan:
Hi @AndrewYakimets. 

Thanks for your interest in `loguru` and thanks for offering enhancing it. Of course, I would happily accept any improvement! Code cleanups are welcome. :wink:

AndrewYakimets:
Hi @Delgan , thank you for fast response! Okay, I am starting :smiley:

AndrewYakimets:
Hi, @Delgan ! I've finished implementing my proposed changes in PR #354. The CI checks were passed. Could you, please, check it?

Delgan:
Checked and merged, thanks!

PR #354, opened 2020-10-23
Fixed code style and flake8 errors.

AndrewYakimets:
As proposed in #353, this PR consist of:

**Fixed _flake8_ errors in tests/, docs/ and setup.py. For example:**
- Unused imports;
- Unused variables and lines of code in tests;
- [Usage of lambda in tests instead of function](https://www.flake8rules.com/rules/E731.html);
- Line length exceeding limit in tests;
- [Usage of bare except](https://www.flake8rules.com/rules/E722.html);
- [Inappropriate comparison of boolean types in tests](https://www.flake8rules.com/rules/E712.html).

**Optimized __Logger_ initialization reading using key/value.** I think, such style will give developers better visual understanding.

Also, **updated the CHANGELOG.rst** as mentioned in [documentation](https://loguru.readthedocs.io/en/stable/project/contributing.html#implementing-changes).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,8 +1,7 @@
-`0.5.4`_ (2020-10-24)
-=====================
+`_Unreleased`_
+==============
 
-- Fix flake8 errors in tests/, docs/ and setup.py (`#353 <https://github.com/Delgan/loguru/issues/353>`_).
-- Optimize _Logger initialization reading using key/value style (`#353 <https://github.com/Delgan/loguru/issues/353>`_).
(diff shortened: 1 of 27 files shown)


Delgan:
Thank you for taking the time to do all this! It must not have been very pleasant, sorry. :grin:

Now that the entire code base has been fixed, it has allowed to add checks for `tests/` to the `flake8` command in `tox.ini`.
I hope this will help to keep these files clean in the future.

I made very little modifications to your PR (excluded files in exception/source, changed "0.5.4" to "Unreleased" as I don't know yet what will be the next release, fixed few remaining lines too long warnings, reworded changelog because "optimize" implies performances in my opinion).

Anyway, thanks again for your dedication to this work!

PR #381, opened 2020-12-31
Add recipe Implement RotatingFileHandler provided by logging package

k8uwall:
I wrote RotatingFileHandler using callable of retention.

It will be rotated like this
file.log file.log.1 file.log.2 file.log.3

### DIFF
--- a/docs/resources/recipes.rst
+++ b/docs/resources/recipes.rst
@@ -370,6 +370,32 @@ The ``rotation`` argument of file sinks accept size or time limits but not both 
     logger.add("file.log", rotation=rotator.should_rotate)
 
 
+Implement RotatingFileHandler provided by logging package
+---------------------------------------------------------
+
+Rotate log files by assigning numbers to them::
+
+    class Retentor:
+        def __init__(self, path, backup_count):
+            self._path = path
+            self._backup_count = backup_count
+
+        def should_retente(self, logs):
+            new_log = sorted(logs).pop(0)
+
+            for i in range(self._backup_count, 0, -1):
+                new = f"{self._path}.{i}"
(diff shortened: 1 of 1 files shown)


Delgan:
Hey @k8uwall, thanks for the PR! This is a good way to use a callable for `retention`.

Can you please briefly explain what are the advantages of using `Retentor` instead of `RotatingFileHandler`?

```python
handler = RotatingFileHandler("file.log", maxBytes=5000, backupCount=3)
logger.add(handler)
```

Also, I may be wrong, but shouldn't you call `os.remove()` somewhere to delete old log files?

k8uwall:
Thank you for your answer.
Very good recipe for those who are thinking of switching from the default logging.
Why not put this in the README.md at the top?


It will be overwritten in this way, so there is no need to delete it in reverse.
```
cp file.log.2 file.log.3
cp file.log.1 file.log.2
cp file.2021-01-08_23-01-38_246557.log file.log.1
```

Delgan:
> It will be overwritten in this way, so there is no need to delete it in reverse.

Oh, yeah, makes sense. Thanks.



> Thank you for your answer.
> Very good recipe for those who are thinking of switching from the default logging.
> Why not put this in the README.md at the top?

Well, I prefer to make the Readme focused on the use of Loguru and not on standard logging. But indeed, I think I could add the `RotatingFileHandler` in the documentation. Just before your recipe for example, or somewhere in the page about [migrating from standard logging](https://loguru.readthedocs.io/en/stable/resources/migration.html).

k8uwall:
Thank you.
I'll send a pull request to the migration page next time.

PR #392, opened 2021-01-31
Update readme

agn-7:
minor update in Readme: PyPy to PyPI, because PyPy is a totally different thing.


Delgan:
Hi @agn-7. :)

Actually, this is really "PyPy" which is intended. The badge is meant to indicate that Loguru supports CPython 3.5 (at least) and also the PyPy interpreter.

agn-7:
So why does refer to PyPI website? I mean was the intended logo

agn-7:
@Delgan Hi
OK, my bad

PR #403, opened 2021-02-19
Add file sink permissions

obviousrebel:
Adds a way to set permissions for files created by file handler.

Potentially should be adapted to instead take int of permissions and convert to octal needed for `os.chmod()`.


Delgan:
I finally took some time to review and update this PR. :grin:

Thanks for your work!

PR #406, opened 2021-02-25
Add logger.catch(default=...) to type hints

gregmuellegger:
Hi there, my IDE shows errors when using the `default` param of `logger.catch`:

grafik

This PR fixes it. Haven't worked with pyi files before, so hope this is the correct way to specify it.
Please shout out if this requires tests of any kind.


Delgan:
I missed that when adding `default` argument. Thanks for having caught that!

PR #496, opened 2021-08-30
Feature/onerror with self

bro-adm:
this allows the onerror method to receive the self, cls on methods that get them -> by using in the onerror methods the key word obj_bound -> but to not break earlier code I made it so that if the method does not have obj_bound in signature it won't pass the self/cls over. in regular and static methods it doesn't pass it ever and raises error if it is requested.

### DIFF
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -1213,7 +1213,7 @@ class Logger:
                 # basically just runs over members of arg_0 and sees if the decorated method is one in one of the tuples
                 # tuple -> (name, val)
                 if args:
-                    if list(filter(lambda attr_tup: True if attr_tup[0] == decorated.__name__ else False, getmembers(args[0])):
+                    if list(filter(lambda attr_tup: True if attr_tup[0] == decorated.__name__ else False, getmembers(args[0]))):
                         self_._decorated_method_bound = args[0]
 
             def __call__(_, function):
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -1169,7 +1169,7 @@ class Logger:
         class Catcher:
             def __init__(self_, from_decorator):
                 self_._from_decorator = from_decorator
-                self_._decorating_method = False
+                self_._decorated_method_Self = None
 
             def __enter__(self_):
                 return None
@@ -1199,7 +1199,7 @@ class Logger:
 
                 return not reraise
 
-            def __check_method_or_func(self, args, decorated):
+            def __method_or_func(self, args, decorated):
                 # return True if the method decorated is a method of cls and false if it does not receive it
                 # class/instance methods -> return True
                 # regular function and static -> return False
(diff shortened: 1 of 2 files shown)


Delgan:
Hey @bro-adm, thanks for wanting to contribute to Loguru.

I understand your use case but honestly I'm not very happy with the idea of having to inspect the signature of the `onerror` function and dynamically adjusting the number of arguments passed to it. Especially if this implies to use a "magic" argument name like `obj_bound`.

This is an advanced solution that I would not have thought of, but it is too fragile in my opinion.

I would prefer that we find a workaround with the means offered by Python that is easily understandable.

If one need the `onerror` to access the class instance, I would suggest to simply decorate the method during initialization:

```python
from loguru import logger

class Foo:

    def __init__(self):
        self.to_decorate = logger.catch(onerror=self._on_error)(self.to_decorate)

    def to_decorate(self):
        raise ValueError("Oops...")

    def _on_error(self, value):
        print(f"Error caught by Loguru in instance {self}")

foo = Foo()
foo.to_decorate()
```


If you find this too cumbersome, you can create your own decorator as follow:

```python
from loguru import logger

def self_catch(*, onerror=lambda self, err: None, **catch_kwargs):

    def decorator(meth):
        def decorated(self, *meth_args, **meth_kwargs):
            with logger.catch(onerror=lambda err: onerror(self, err), **catch_kwargs):
                return meth(self, *meth_args, **meth_kwargs)
        return decorated

    return decorator

class Foo:

    @self_catch(onerror=lambda self, err: self._on_error(err))
    def to_decorate(self):
        raise ValueError("Oops...")

    def _on_error(self, value):
        print(f"Error caught by Loguru in instance {self}")

foo = Foo()
foo.to_decorate()
```

bro-adm:
honestly I get it, I wasn't to happy with it myself and therefore thought to about the possibility to have on error receive all the method decorated params -> *args, **kwargs and by doing that it would allow the self/cls and all other params to be passed on for the coder to decide the use case for it. but I might not be understanding correctly the use case of the onerror method -> what do people usually do with it?

bro-adm:
is it used for prints, extra logging, instance status changes...?

Delgan:
The main use case for `onerror` I think is to call `sys.exit(1)` while decorating the `main()` function. That way:
* unexpected error will be logged and displayed thanks to `@catch()`
* the program will return a non-zero error indicating a failure
*  immediate exit means the exception won't be duplicated on the terminal if we added `reraise=True`. 

However, it can be used in any way. The idea of this argument is to give the user the possibility to handle the errors as he needs.

bro-adm:
how do you use the decorator as a with statement --> i tried to look it up but i did not find anything like what you did only a workaround

Delgan:
Which decorator do you mean? Maybe can you use [`@contextlib.contextmanager`](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager)?

bro-adm:
thank you

Issue #372, opened 2020-11-30
Allow custom serializers


XayOn:
The hability to specify a custom serializer for a file could be useful to save files in human-readable serializable formats such as the one followed by splunk. This would also open a door to save pickled, xml or any other log format the user wants (for whatever reason... I cant think of a good one to save pickled data or xml in logs tbh).

Should be simple enough. I'm proposing a backwards-compatible implementation where serializer is either a boolean, or a custom serializer directly.

https://github.com/Delgan/loguru/blob/1b41299643fbab269f89878226211a24bdb3e6b0/loguru/_handler.py#L258

XayOn:
I'm leaving this idea here, but will come back tomorrow with a pull request for this proposal 😃

Delgan:
Hi! Thanks for your interest on `loguru` and thanks for being willing to contribute. 

This feature has been requested several times but so far I have always been reluctant to add it.
The most recent discussion on this topic: https://github.com/Delgan/loguru/issues/339#issuecomment-702013972

I don't know if you saw it but there exists a snippet in the documentation: [Serializing log messages using a custom function](https://loguru.readthedocs.io/en/latest/resources/recipes.html#serializing-log-messages-using-a-custom-function)

I think using a custom `formatter` is not too cumbersome and should be the way to go. I prefer suggesting this solution over complicating the `serialize` parameter.

```python
def formatter(record):
    record["extra"]["serialized"] = orjson.dumps(record, default=str)
    return "{extra[serialized]}\n"

logger.add("file.log", format=formatter)
```

In this way, the user has the possibility to customize the serialization process entirely.

XayOn:
The patch() function is actually better for what I was looking for!
I thought I read enough on the documentation but I missed that part.

Thanks and thanks for this great library!

Delgan:
Thanks for your understanding, @XayOn!

Issue #339, opened 2020-09-29
Make default file sink encoding utf8


jessekrubin:
Making the default file encoding utf-8 would be quite nice (in my opinion and would be better for JSON serialization).

Delgan:
Well, I tried to copy the default behavior of `open()` which uses the encoding returned by [`locate.getpreferredencoding()`](https://docs.python.org/3/library/locale.html#locale.getpreferredencoding). It usually returns `"utf-8"` on Linux and MacOS, but not necessarily on Windows.

There actually exists a big discussion about the question of default locale encoding for Python files: [PEP 597: Use UTF-8 for default text file encoding](https://discuss.python.org/t/pep-597-use-utf-8-for-default-text-file-encoding/1819/47).

It seems reasonable to make `loguru` uses `"utf8"` encoding by default regardless of the platform. I will first need to review the discussion I linked above to make sure there are no possible issues.

As you mentioned JSON serialization, I also think we should set `ensure_ascii=False` here: https://github.com/Delgan/loguru/blob/93f199042c5612367aab0cbee018961b1748627b/loguru/_handler.py#L258

Otherwise, non-ascii characters are escaped despite the file being opened with `"utf-8"` encoding.

jessekrubin:
I think you kinda have made the decision already tho by including emoji-icon things that don't render on my windows machine. It also creates technically invalid JSON (tho it works with the stdlib json parser/serializer).

I have several scripts that use loguru but monkey patch in a new `def _serialize_record(text, record):` that uses orjson/python-rapidjson. Both libs are faster than the stdlib json-module AND are stricter about sticking to creating valid JSON. The only thing I have to do is set the encoding of the file sink to make this work on windows. 

The PEP discussion you link to is pretty interesting, but I don't think the reasons they mention for not making the default encoding utf8 (that some windows users just want to do their job and not delve into the oddities of file encodings) apply to the average lo-guru. 

Unrelated: I was thinking of opening a pull request if/when I find the time to allow the serialize option for a sink to be either a bool or a callable that accepts `(test, record)` and returns a string.

Delgan:
> I think you kinda have made the decision already tho by including emoji-icon things that don't render on my windows machine.

Fair point. :smile:
Although such symbols were not mean to be displayed if sink does not support unicode, be it a file or the console. That's why there is no `{level.icon}` in the default format. However, it is true that if `serialize=True` there is no choice anyway.

[...]

Issue #278, opened 2020-06-05
Accept a Callable for the serialize argument when adding a sink


gjeusel:
Hello Delgan ! Love your library !

I was wondering what you would think about accepting a Callable for the `serialize` parameter in `logger.add` method ?

This would ease customization of the jsonified records. And I'm thinking about ELK stack usage for example.

I managed to achieve this purpose by overwriting the `Handler._serialize_record` staticmethod:

```python
import sys

import flask
import loguru
import orjson
import pytz
import stackprinter
from flask_login import current_user


def _serialize_record_elk(text, record):
    # Format timestamp for Elasticache
    timestamp = record["time"].astimezone(pytz.utc)
    timestamp = timestamp.strftime("%Y-%m-%dT%H:%M:%S.{:03d}Z").format(
        timestamp.microsecond // 1000
    )

    # Current User's email of flask app
    user_email = getattr(current_user, "email", "anonymous")

    # Can add some config variable of interest from the flask app
    env = None
    if flask.has_request_context():
        if hasattr(flask.current_app, "config"):
            env = flask.current_app.config["ENV"]

    serializable = {
        # Base fields
        "@timestamp": timestamp,  # Elasticache index overwrite
        "icon": record["level"].icon,
        "log": {"level": record["level"].name},
        "name": record["name"],
        "message": record["message"],
        # App related:
        "user": {"email": user_email},
        "env": env,
        # Extra
        "extra": record["extra"],
        "misc": {
            "elapsed": {
                "repr": record["elapsed"],
                "seconds": record["elapsed"].total_seconds(),
            },
            "file": {"name": record["file"].name, "path": record["file"].path},
            "function": record["function"],
            "level": {
                "icon": record["level"].icon,
                "name": record["level"].name,
                "no": record["level"].no,
            },
            "line": record["line"],
            "module": record["module"],
            "process": {"id": record["process"].id, "name": record["process"].name},
            "thread": {"id": record["thread"].id, "name": record["thread"].name},
        },
    }
    if record["exception"]:
        exc_class = record["exception"].type
        serializable = {
            **serializable,
            "exc": {
                "type": f"{exc_class.__module__}.{exc_class.__name__}",
                "value": record["exception"].value,
                "traceback": stackprinter.format(record["exception"], show_vals="line"),
            },
        }

    return orjson.dumps(serializable, default=str).decode("utf-8") + "\n"


loguru._handler.Handler._serialize_record = staticmethod(_serialize_record_elk)

[...]

Issue #203, opened 2020-01-17
Would a custom json format in serialize_record have any value for loguru?


devspyrosv:
Hello!
There are cases where the extra arguments would fit nicely at the top level of the json dict.

By having the json format hardcoded in 'serialize_record' as 'serializable' it's not possible to change the final json format.

I believe that the 'serializable' variable should be accessible  just like 'format' is when adding a logger.
Would you agree with the above?

Delgan:
Hi! :)

Extending the `serialize` attribute so that it accepts a function is a possibility, but if possible I'd rather avoid that. I prefer arguments being simple `bool` values, it's easier to reason about.

Basically, I think that if someone needs a custom way to handle logging messages, it should be implemented in a custom sink. This is the place where you transform a record to an appropriate representation.

For now, I see two "easy" ways to customize the serialization process: either add extra attributes with `bind()` /  `patch()` so that they are automatically serialized or simply use `json.dumps()` in your own sink. Is your use case not compatible with these solutions?

devspyrosv:
I just thought that having the json format accessible just like the non serializable formatting is, would be powerful. 

Using bind and patch breaks the compatibility with the native logger and since loguru has the ability to be used similarly to it (for simple cases as a drop in replacement) I would prefer not to use them in large projects.

Checking out json.dumps in a custom sink seems right.
Thank you!

Delgan:
> I just thought that having the json format accessible just like the non serializable formatting is, would be powerful.

I do think there are cases where it could be useful. But I also have to consider whether it's worth making the api more complex, compared to existing solutions.

> Using bind and patch breaks the compatibility with the native logger and since loguru has the ability to be used similarly to it (for simple cases as a drop in replacement) I would prefer not to use them in large projects.

I understand. But in this case, it is difficult for me to justify extending the api when there already exists a possible solution. Maybe you can use the `extra` / `patcher` attributes of the `logger.configure()` method? If you need to go back to standard `logging`, that's just one line to remove.


> Checking out json.dumps in a custom sink seems right.

Note though that as Loguru uses some custom types, you will probably need to use a fall-back serializer like `json.dumps(record, default=str)`.

devspyrosv:
Ok! Thank you for the hint about the fall-back serializer!

Delgan:
No problem. :)

If you have any other questions or suggestions, don't hesitate to let me know. :+1:

mrkovalchuk:
Hello!

[...]

PR #521, opened 2021-10-14
Custom serializer

sripathivenky:
* adding serializer that allows users to pass a  method to work in hand with serialize flag.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -282,6 +282,52 @@ Want your logs to be serialized for easier parsing or to pass them around? Using
 
     logger.add(custom_sink_function, serialize=True)
 
+Pass in custom serializer method to get more custom JSON out.
+
+::
+
+    def custom_serializer(formated_text: any, record: dict) -> str:
+        exception = record["exception"]
+
+        if exception is not None:
+            exception = {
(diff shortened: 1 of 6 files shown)


Delgan:
Hi @sripathivenky.

Sorry for being late to answer and thanks for taking of your time to improve Loguru.

The idea to add a new `serializer` parameter has been discussed several times (https://github.com/Delgan/loguru/issues/372#issuecomment-736774742, https://github.com/Delgan/loguru/issues/339#issuecomment-702013972, https://github.com/Delgan/loguru/issues/278#issuecomment-639519441, https://github.com/Delgan/loguru/issues/203#issuecomment-575684228). I'm reticent to do so because I'm not fan of such optional parameter while there already exists solutions which provide the same result: [Serializing log messages using a custom function](https://loguru.readthedocs.io/en/latest/resources/recipes.html#serializing-log-messages-using-a-custom-function).

Using a custom `sink`, `formatter` or even `patch()` allow anyone to customize the serialization without much efforts.

I understand that a `serializer` option would make it more straightforward but it also adds some complexity I prefer to avoid.

PR #572, opened 2022-01-19
Docs for caplog integration that avoids deadlocks

### LINKED ISSUE #474: Pytest and caplog doesn't work if InterceptHandler exists
I would like route all logs through loguru and configure them as I see fit, and for that I have created an `InterceptHandler` as descriped [here](https://loguru.readthedocs.io/en/stable/overview.html?highlight=intercepthandler#entirely-compatible-with-standard-logging).
I also want to test the output of some of my logs via the caplog trick described [here](https://loguru.readthedocs.io/en/stable/resources/migration.html?highlight=pytest#making-things-work-with-pytest-and-caplog).

It seems that I can't have both at the same time, as it creates a hanging loop.

Is there another way to test the contents of the logs when using an `InterceptHandler` ?

phillipuniverse:
The previous suggestion resulted in a deadlock when an InterceptHandler was used to integrate standard logging into loguru. Fixes #474.

### DIFF
--- a/docs/resources/migration.rst
+++ b/docs/resources/migration.rst
@@ -230,22 +230,16 @@ Making things work with Pytest and caplog
 
 If you've followed all the migration guidelines thus far, you'll notice that this test will fail. This is because |pytest|_ links to the standard library's ``logging`` module.
 
-So to fix things, we need to add a sink that propagates Loguru to ``logging``.
-This is done on the fixture itself by mokeypatching |caplog|_. In your ``conftest.py`` file, add the following::
+So to fix things, we need to add a sink that propagates Loguru to the caplog handler.
+This is done by overriding the |caplog|_ fixture to capture its handler. In your ``conftest.py`` file, add the following::
 
-    import logging
     import pytest
-    from _pytest.logging import caplog as _caplog
     from loguru import logger
 
     @pytest.fixture
     def caplog(_caplog):
-        class PropogateHandler(logging.Handler):
-            def emit(self, record):
-                logging.getLogger(record.name).handle(record)
-
-        handler_id = logger.add(PropogateHandler(), format="{message} {extra}", level="TRACE")
+        handler_id = logger.add(_caplog.handler, format="{message}")
         yield _caplog
         logger.remove(handler_id)
 
-Run your tests and things should all be working as expected. Additional information can be found in `GH#59`_.
+Run your tests and things should all be working as expected. Additional information can be found in `GH#59`_ and `GH#474`_.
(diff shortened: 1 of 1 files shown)


docs/resources/migration.rst
```diff
@@ -230,22 +230,16 @@ Making things work with Pytest and caplog
...
 If you've followed all the migration guidelines thus far, you'll notice that this test will fail. This is because |pytest|_ links to the standard library's ``logging`` module.
 
-So to fix things, we need to add a sink that propagates Loguru to ``logging``.
-This is done on the fixture itself by mokeypatching |caplog|_. In your ``conftest.py`` file, add the following::
+So to fix things, we need to add a sink that propagates Loguru to the caplog handler.
+This is done by overriding the |caplog|_ fixture to capture its handler. In your ``conftest.py`` file, add the following::
 
-    import logging
     import pytest
-    from _pytest.logging import caplog as _caplog
```
phillipuniverse, line 238: I didn't think this was right. I believe this works because Pytest injects `_caplog` before exporting a new `caplog` fixture (the one we define below). In fact, I think even before this import didn't actually do anything.

Delgan: You're probably right. The issue is that it generates a `fixture '_caplog' not found` error.

What do you think of overriding the existing `caplog` directly?

```python
@pytest.fixture
def caplog(caplog):
    handler_id = logger.add(caplog.handler, format="{message}")
    yield caplog
    logger.remove(handler_id)
```

docs/resources/migration.rst
```diff
@@ -230,22 +230,16 @@ Making things work with Pytest and caplog
...
-            def emit(self, record):
-                logging.getLogger(record.name).handle(record)
-
-        handler_id = logger.add(PropogateHandler(), format="{message} {extra}", level="TRACE")
+        handler_id = logger.add(_caplog.handler, format="{message}")
         yield _caplog
         logger.remove(handler_id)
 
-Run your tests and things should all be working as expected. Additional information can be found in `GH#59`_.
+Run your tests and things should all be working as expected. Additional information can be found in `GH#59`_ and `GH#474`_.
```
Delgan, line 245: Can you please reference the link at the very top of the `.rst` document here (otherwise it won't work): https://github.com/Delgan/loguru/blame/f6b6a54f3e0090dba22e8246a189fc68945968ca/docs/resources/migration.rst#L45 ?

Delgan (changes requested):

docs/resources/migration.rst
phillipuniverse: For some reason I thought the way to override system-level fixtures was to prefix with an underscore. I now see that isn't right - should just request the fixture itself and simply re-export it.

I also did you 1 better and type-hinted what `caplog` actually is 😉 

I tested in my project and it works great!

Delgan (approved):

Delgan:
Great, thanks a lot for your time! :tada:

PR #694, opened 2022-08-19
feature: Use exceptions consistently instead of record['exception']

ponponon:
Use exceptions consistently instead of record['exception']


Delgan:
Great, thanks for noticing that!

PR #705, opened 2022-09-14
Fix log injection

pim-hoeven:
This PR fixes the log injection issue as described [here](https://www.huntr.dev/bounties/73ebb08a-0415-41be-b9b0-0cea067f6771/). The issue got attention in [bc1dab](https://www.github.com/delgan/loguru/commit/bc1dab4cc5044d704a980f5675b9d02f9f61b528), but was never fixed, so our repository manager was [still complaining](https://ossindex.sonatype.org/vulnerability/sonatype-2022-0703?component-type=pypi&component-name=loguru):

image

Happy to hear your thoughts.

### DIFF
--- a/docs/resources/recipes.rst
+++ b/docs/resources/recipes.rst
@@ -93,7 +93,7 @@ You should also avoid logging a message that could be maliciously hand-crafted b
     logger.info(message, value=SomeValue(10))
 
 
-Another danger due to external input is the possibility of a log injection attack. Consider that you may need to escape user values before logging them: `Is your Python code vulnerable to log injection? <https://dev.arie.bovenberg.net/blog/is-your-python-code-vulnerable-to-log-injection/>`_
+Another danger due to external input is the possibility of a log injection attack. This can happen with the option `raw=True`. Consider that you may need to escape user values before logging them: `Is your Python code vulnerable to log injection? <https://dev.arie.bovenberg.net/blog/is-your-python-code-vulnerable-to-log-injection/>`_
 
 .. code::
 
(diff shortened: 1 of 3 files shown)


pim-hoeven:
@Delgan let me know if there's anything I need to improve on. Thanks!

Delgan:
Hey @pim-hoeven.

Sorry for the late answer and thank you for taking the time to open a PR. :+1: 

I indeed referenced the log injection attack in the documentation but didn't add escaping to the `logger` implementation. The reason being that it is common for the user to want to log a message on multiple lines, and I don't want to remove this feature by default.

It is however possible to configure your application so that it escapes new lines from messages:

```python
import sys
from loguru import logger

def remove_newline(record):
    escaped = str(record["message"]).replace("\n", "\\n").replace("\r", "\\r")
    record["extra"]["escaped_message"] = escaped

logger.configure(patcher=remove_newline)

logger.remove()
logger.add(sys.stderr, format="[{time}][{level}] {extra[escaped_message]}")

logger.info("Multiple\nlines\nare\nescaped")
```

There is a lot of risks mixing user inputs with logs and it's very hard for Loguru to even mitigating them without affecting usability.

In my opinion, it would be better to analyze the user code and post a security warning only where the logger is called with unsafe arguments. I though about creating CodeQL queries for this but I have no idea how feasible it is.

pim-hoeven:
Hi @Delgan, thanks for taking the time to look at this PR and providing me with an example! Unfortunately, this does not fix the vulnerability in general, so we will not be able to use the package anymore because the default would still be 'vulnerable'. I understand your reasoning from a usability perspective, so I'll close the PR. I really enjoyed using loguru, good luck with developing it further!

PR #713, opened 2022-09-27
Fix deadlock caused by generational GC and object finalizers

### LINKED ISSUE #712: Deadlock when objects emit messages at garbage collection
Hello :wave: 

I have found a issue whereby loguru can hit a deadlock due to garbage collection. Specifically, if there is an object in memory that has the following properties:

1. Is part of a [Cyclic Isolate](https://legacy.python.org/dev/peps/pep-0442/#id2) (i.e. it will be garbage collected by the generational garbage collector, _not_ the reference counter)
2. Implements a `__del__` finalizer which explicitly or implicitly triggers a log statement

Then any memory assignment during log handling has a chance to trigger a deadlock.

## Reproducible test case

**Python version**: 3.8.12
**Loguru version**: 0.6.0

The following is a script which demonstrates the deadlock:

```python
import sys

import loguru


class CyclicReference:
    """A minimal cyclic reference.

    Cyclical references are garbage collected using the generational collector rather than
    via reference counting. This is important here, because the generational collector runs
    periodically, meaning that it is hard to predict when the stack will be overtaken by a
    garbage collection process - but it will almost always be when allocating memory of some
    kind.

    When this object is garbage-collected, a log will be emitted.
    """
    def __init__(self, _other: "CyclicReference" = None):
        self.other = _other or CyclicReference(_other=self)

    def __del__(self):
        loguru.logger.info("Goodbye!")


class Sink:
    def write(self, message):
        # This simulates assigning some memory as part of log handling. Eventually this will
        # trigger the generational garbage collector to take over here.
        # You could replace it with `json.dumps(message.record)` to get the same effect.
        _ = [dict() for _ in range(20)]

        # Actually write the message somewhere
        sys.stderr.write(message)

    def flush(self):
        sys.stderr.flush()


def main():
    # Set up logger with sink:
    loguru.logger.add(Sink(), level="INFO")

    # Keep logging until a cyclic reference is garbage collected, at which point deadlock
    # should be reached:
    for index in range(10000):
        CyclicReference()  # Here we create cyclic isolates, simulating code which uses them
        loguru.logger.info("Hello", index=index)


if __name__ == "__main__":
    main()

```

### Expected behaviour

A series of `Hello` logs are emitted, interspersed with `Goodbye!` logs.

### Actual behaviour

The program reaches a deadlock, requiring a SIGTERM to exit.

## Explanation of the issue

The `emit` method of loguru's handler [uses a lock](https://github.com/Delgan/loguru/blob/3d8631ef0015267ad68c5d65a250cb1aaa0b61fb/loguru/_handler.py#L171) to avoid contention for the log handler. It is for this reason, that emitting additional log messages in custom sinks is not supported and will result in deadlocks (see #390).

This seems to be a relatively sensible limitation, however the nature of the generational garbage collector in CPython can cause this to happen inadvertently and unpredictably in code which has cyclic isolates and performs some level of processing during log handling.

## Real world example

The reproducible example above seems contrived, and one might suggest that the answer is simply to avoid reference cycles, so I will give an actual example of this that I've seen. The [Python Elasticsearch client](https://github.com/elastic/elasticsearch-py) contains [reference cycles](https://github.com/elastic/elasticsearch-py/blob/v7.17.6/elasticsearch/_async/client/__init__.py#L190-L243) as a means for sharing connections between interfaces.

As a transport it uses [aiohttp](https://github.com/aio-libs/aiohttp), which implements a [`__del__` garbage collection finalizer](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L295-L306). This in turn invokes the asyncio exception handler, which [defaults to emitting a log statement](https://github.com/python/cpython/blob/a0ad63e70e3682cdf7e87e28091bb54fe12a2d4e/Lib/asyncio/base_events.py#L1760). Since this is a stdlib `logging` call, it doesn't cause a problem on its own, but it is likely that users of loguru will set up some kind of intercept handler to forward third party stdlib logs to `loguru`.

## Additional notes

It took me some time to figure out a reproducible example for this, as I was only familiar with the reference counting part of the garbage collector. I found the [this article](https://rushter.com/blog/python-garbage-collector/) useful to understand the generational collector, which is what led me to isolate the circumstances to reproduce.

jacksmith15:
This PR aims to solve #712 by using re-entrant locks on the logging handler. This allows the same thread to acquire the lock multiple times. This appears to resolve deadlocks when the generational (or cyclic) garbage collector triggers a emits a log during the handling of another log message.

I'm not clear on whether using a re-entrant lock here allows non-safe situations in other cases, but the existing threading and multi-processing tests do all pass.

Resolves #712

### DIFF
--- a/docs/resources/recipes.rst
+++ b/docs/resources/recipes.rst
@@ -145,7 +145,7 @@ The most straightforward workaround is to |remove| your handler and then re-|add
 
     ...
 
-    logger.remove(handler_id)
+    logger.remove(handler_id)  # For the default handler, it's actually '0'.
     logger.add(sys.stderr, level="DEBUG")
 
     logger.debug("Logging 'DEBUG' messages too")


tests/test_locks.py
```diff
@@ -0,0 +1,50 @@
...
+    def __del__(self):
+        logger.info("tearing down")
+
+
+def test_no_deadlock_on_generational_garbage_collection():
+    """Regression test for https://github.com/Delgan/loguru/issues/712
+
+    Assert that deadlocks do not occur when a cyclic isolate containing log output in
+    finalizers is collected by generational GC, during the output of another log message.
+    """
```
jacksmith15, line 27: I'm not particularly happy with the test, as when it fails it obviously results in deadlock rather than passing or failing, and the test process requires a SIGTERM to exit.

This could be worked around by invoking a subprocess and timing out - but that requires a fair bit of scaffolding, and makes the code harder to grok. Happy to do this if its preferred.

tests/test_locks.py
```diff
@@ -0,0 +1,50 @@
...
+        _ = [dict() for _ in range(20)]
+
+        # Actually write the message somewhere
+        output.append(message)
+
+    logger.add(sink, colorize=False)
+
+    # WHEN there are cyclic isolates in memory which log on GC
+    # AND logs are produced long enough to trigger generational GC
+    for _ in range(1000):
```
jacksmith15, line 45: This number is somewhat arbitrary unfortunately. The generational GC will trigger after a certain amount of memory is allocated. Possibly this depends on the amount of memory available? On my machine `range(10)` did not reproduce the issue, but `range(100)` did. I went for `range(1000)` to be safe. Its still quite quick - 130ms on my machine for the entire test.

jacksmith15:
I believe the failing tests all relate to an implicit upgrade of `mypy` from 0.971 to 0.981 (the version is unpinned). They haven't made release notes available for v0.981 yet, but the diff is [here](https://github.com/python/mypy/compare/v0.971...v0.981).

I will push a commit to ensure mypy runs with the same version as it did for the most recent commit on `master`.

jacksmith15:
The remaining failing test appears to be a network failure uploading the coverage report. I can't re-run it, and I don't want to push a commit that re-runs _all_ of the tests unnecessarily.

jacksmith15:
Closing due to issues with re-entrant locks as explained here: https://github.com/Delgan/loguru/issues/712#issuecomment-1271489577

PR #720, opened 2022-10-07
Avoid deadlock caused by generational GC and object finalizers

### LINKED ISSUE #712: Deadlock when objects emit messages at garbage collection

jacksmith15:
This PR implements a solution to #712 as discussed [here](https://github.com/Delgan/loguru/issues/712#issuecomment-1263719117).

If a re-entrant logging call is performed (a particular thread emits a message whilst in the process of emitting another), then the log will not be emitted, and instead handled by the error interceptor. This differs from the current behaviour, which is that a deadlock would be reached.

Resolves #712

### DIFF
--- a/loguru/_handler.py
+++ b/loguru/_handler.py
@@ -104,9 +104,9 @@ class Handler:
         if getattr(self._lock_acquired, "acquired", False):
             raise RuntimeError(
                 "Could not acquire internal lock because it was already in use (deadlock avoided). "
-                "This likely happened because the logger was re-used inside a sink, a signal handler "
-                "or a '__del__' method. This is not permitted because the logger and its handlers are "
-                "not re-entrant."
+                "This likely happened because the logger was re-used inside a sink, a signal "
+                "handler or a '__del__' method. This is not permitted because the logger and its "
+                "handlers are not re-entrant."
             )
         self._lock_acquired.acquired = True
         try:
--- a/loguru/_handler.py
+++ b/loguru/_handler.py
@@ -317,6 +317,7 @@ class Handler:
     def __getstate__(self):
         state = self.__dict__.copy()
         state["_lock"] = None
+        state["_lock_acquired"] = None
         state["_memoize_dynamic_format"] = None
         if self._enqueue:
             state["_sink"] = None
@@ -327,6 +328,7 @@ class Handler:
     def __setstate__(self, state):
         self.__dict__.update(state)
         self._lock = create_handler_lock()
+        self._lock_acquired = threading.local()
         if self._is_formatter_dynamic:
             if self._colorize:
                 self._memoize_dynamic_format = memoize(prepare_colored_format)
--- a/loguru/_handler.py
+++ b/loguru/_handler.py
@@ -5,7 +5,7 @@ import os
 from threading import Thread
 
 from ._colorizer import Colorizer
-from ._locks_machinery import create_handler_lock
+from ._locks_machinery import HandlerLockNotReentrant, create_handler_lock
 
 
 def prepare_colored_format(format_, ansi_level):
@@ -175,7 +175,8 @@ class Handler:
                     self._queue.put(str_record)
                 else:
                     self._sink.write(str_record)
-
+        except HandlerLockNotReentrant:
+            self._error_interceptor.print(record)
         except Exception:
             if not self._error_interceptor.should_catch():
                 raise
--- a/loguru/_handler.py
+++ b/loguru/_handler.py
@@ -2,10 +2,12 @@ import functools
 import json
 import multiprocessing
 import os
+import threading
+from contextlib import contextmanager
 from threading import Thread
 
 from ._colorizer import Colorizer
-from ._locks_machinery import HandlerLockNotReentrant, create_handler_lock
+from ._locks_machinery import create_handler_lock
 
 
 def prepare_colored_format(format_, ansi_level):
@@ -64,6 +66,7 @@ class Handler:
 
         self._stopped = False
         self._lock = create_handler_lock()
+        self._lock_acquired = threading.local()
         self._queue = None
         self._confirmation_event = None
         self._confirmation_lock = None
@@ -94,6 +97,24 @@ class Handler:
 
     def __repr__(self):
         return "(id=%d, level=%d, sink=%s)" % (self._id, self._levelno, self._name)
+
+    @contextmanager
+    def _protected_lock(self):
+        """Acquire the lock, but fail fast if its already acquired by the current thread."""
+        if getattr(self._lock_acquired, "acquired", False):
+            raise RuntimeError(
+                "Could not acquire internal lock because it was already in use (deadlock avoided). "
+                "This likely happened because the logger was re-used inside a sink, a signal handler "
+                "or a '__del__' method. This is not permitted because the logger and its handlers are "
+                "not re-entrant."
+            )
+        self._lock_acquired.acquired = True
+        try:
+            self._lock.acquire()
+            yield
+        finally:
+            self._lock.release()
+            self._lock_acquired.acquired = False
 
     def emit(self, record, level_id, from_decorator, is_raw, colored_message):
         try:
@@ -168,22 +189,20 @@ class Handler:
             str_record = Message(formatted)
             str_record.record = record
 
-            with self._lock:
+            with self._protected_lock():
                 if self._stopped:
                     return
                 if self._enqueue:
                     self._queue.put(str_record)
                 else:
                     self._sink.write(str_record)
-        except HandlerLockNotReentrant:
-            self._error_interceptor.print(record)
         except Exception:
             if not self._error_interceptor.should_catch():
                 raise
             self._error_interceptor.print(record)
 
     def stop(self):
-        with self._lock:
+        with self._protected_lock():
             self._stopped = True
             if self._enqueue:
                 if self._owner_process_pid != os.getpid():
@@ -208,7 +227,7 @@ class Handler:
         if self._enqueue and self._owner_process_pid != os.getpid():
             return
 
-        with self._lock:
+        with self._protected_lock():
             await self._sink.complete()
 
     def update_format(self, level_id):
--- a/loguru/_locks_machinery.py
+++ b/loguru/_locks_machinery.py
@@ -2,13 +2,44 @@ import os
 import threading
 import weakref
 
+
+class HandlerLockNotReentrant(RuntimeError):
+    pass
+
+
+class _HandlerLockWrapper:
+    """Lock wrapper for the handler.
+
+    This avoids deadlock by propagating an exception rather than waiting forever.
+    """
+
+    def __init__(self, lock):
+        self._lock = lock
+        self._local = threading.local()
+        self._local.acquired = False
+
+    def __enter__(self):
+        if getattr(self._local, "acquired", False):
+            raise HandlerLockNotReentrant(
+                "Tried to emit a log in the process of handling a log. Log handler is not re-entrant."
+            )
+        self._local.acquired = True
+        self._lock.acquire()
+        return
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self._lock.release()
+        self._local.acquired = False
+        return
+
+
 if not hasattr(os, "register_at_fork"):
 
     def create_logger_lock():
         return threading.Lock()
 
     def create_handler_lock():
-        return threading.Lock()
+        return _HandlerLockWrapper(threading.Lock())
 
 else:
     # While forking, we need to sanitize all locks to make sure the child process doesn't run into
@@ -47,4 +78,4 @@ else:
     def create_handler_lock():
         lock = threading.Lock()
         handler_locks.add(lock)
-        return lock
+        return _HandlerLockWrapper(lock)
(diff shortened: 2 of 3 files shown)


loguru/_handler.py
```diff
@@ -175,7 +175,8 @@ def emit(self, record, level_id, from_decorator, is_raw, colored_message):
                     self._queue.put(str_record)
                 else:
                     self._sink.write(str_record)
-
+        except HandlerLockNotReentrant:
```
jacksmith15, line 178: I _think_ this shouldn't be re-raised.

If we swallow it here, we don't emit the log, but the object finalizer can complete whatever else it should do. If we raise it, it will likely still be swallowed (as is the case with finalizers), and the finalizer stop.

loguru/_locks_machinery.py
```diff
@@ -2,13 +2,45 @@
...
+
+
 if not hasattr(os, "register_at_fork"):
 
     def create_logger_lock():
         return threading.Lock()
 
     def create_handler_lock():
-        return threading.Lock()
+        return _HandlerLockWrapper(threading.Lock())
```
jacksmith15, line 43: The approach here changes the behaviour for the handler lock in general. So this applies not only to `Handler.emit`, but also to `Handler.stop`, and `Handler.complete_async`.

I'm not sure if that is correct? If it is, do we need additional handling in those methods?

If this behaviour should only apply for `emit`, I think the solution shouldn't touch this file at all, and I will move it to `_handler.py`.

loguru/_handler.py
Delgan: So, I've been wondering about this, and I think we should not catch `HandlerLockNotReentrant` specifically. Instead, it can be handled by the `except Exception:` block bellow with conditional re-raising depending on `catch` argument.

It's true that it can cause `__del__` implementation to fail if the error is re-raised (because it won't be able to complete), but that's the expected behavior. Using the `logger` with `catch=False` means it can fail for different reasons, either because the lock is already in use or because the custom sink contains a bug for example.

Although the `__del__` documentation specifies raised exceptions are ignored by CPython, they do cause a warning to be displayed, so I don't think they will go unnoticed or be difficult to debug.

Finally, I think that this new safety measure will be triggered mostly when a user reuses the logger in a sink. In such case, the application should fail if the user specified `catch=False`.

loguru/_locks_machinery.py
Delgan: Indeed, we should also protect the other methods using the `lock`, this is a very good idea!

However, I'm thinking it might be better to explicitly wrap the `lock` on use instead of on creation. I prefer to differentiate the context manager of the lock itself because in some cases it is not required. In particular, I'm thinking to [the `lock` inside the `_queued_writer` thread](https://github.com/Delgan/loguru/blob/08c9a871c706463a17142aaae90be0fd71e5a51f/loguru/_handler.py#L268) which does not need to be protected.

I'm thinking of replacing `with self._lock` with `with self._protected_lock():` for example. There is certainly a risk if we forget to use it later, but I'll keep an eye on it and I think it should be fine.

loguru/_locks_machinery.py
```diff
@@ -2,13 +2,46 @@
...
+    """
+
+    def __init__(self, lock):
+        self._lock = lock
+        self._local = threading.local()
+        self._local.acquired = False
+
+    def __enter__(self):
+        if getattr(self._local, "acquired", False):
+            raise HandlerLockNotReentrant(
```
Delgan, line 24: Given my other comment about not handling `HandlerLockNotReentrant` specifically, I think we can just raise a `RuntimeError` here.

Also, given that you suggested to protect all methods, I would make the message a bit more generic (because the error won't necessarily be raised when we "tried to emit a log"). For example:

```python
raise RuntimeError(
    "Could not acquire internal lock because it was already in use (deadlock avoided). "
    "This likely happened because the logger was re-used inside a sink, a signal handler "
    "or a '__del__' method. This is not permitted because the logger and its handlers are "
    "not re-entrant."
)
```

loguru/_locks_machinery.py
```diff
@@ -2,13 +2,46 @@
 import threading
 import weakref
 
+
+class HandlerLockNotReentrant(RuntimeError):
+    pass
+
+
+class _HandlerLockWrapper:
```
Delgan, line 10: Can we use [`@contextlib.contextmanager`](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) for simplification?

tests/test_locks.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+
+    # GIVEN a sink which assigns some memory
+    output = []
+
+    def sink(message):
+        # This simulates assigning some memory as part of log handling. Eventually this will
+        # trigger the generational garbage collector to take over here.
+        # You could replace it with `json.dumps(message.record)` to get the same effect.
+        _ = [dict() for _ in range(20)]
```
Delgan, line 51: I actually don't understand why do we need to trigger the garbage collector this way instead of calling `gc.collect()`? Also, do we actually need to the cyclic references? Isn't the problem related to `__del__` alone, meaning calling `del some_object` triggers the deadlock by itself as long as `some_object.__del__` tries to log a message (and that the GC runs instantly)?

Delgan (commented):
Hey @jacksmith15, thanks a lot for your reactivity and willingness to open a new PR!

I have left a few comments. If you don't have time to address them, I can do it myself, no problem, just let me know. :+1:

loguru/_locks_machinery.py
jacksmith15: Yes we can, but only if we separate the Lock from this logic. E.g.

```python
with _acquire_protected(self._lock):
    ...
```

This works given your suggestion to avoid this behaviour on the queued writer.

Another alternative is to keep the wrapper object, and have a contextlib interface like this:

```python
with self._lock.acquire(raise_on_reenter=True):
    ...
```

Although acquire is a poor choice of method name given its meaning on `threading.Lock`, and I can't think of a better name.
```

File: tests/test_locks.py (hunk 5 above)
Comment (jacksmith15) [reply tests/test_locks.py:51]: Functionally that works!

I based this on the example on the issue, and I avoided explicitly calling `gc.collect` there to demonstrate how this could be triggered reliably with seemingly innocent operations here.

In short, I think this demonstrates how easy it would be to trigger GC at this point, for readers who are not familiar with how the generational GC works.

Happy to switch to `gc.collect` if you'd prefer?

File: loguru/_locks_machinery.py (hunk 3 above)
Comment (jacksmith15) [reply loguru/_locks_machinery.py:24]: 👍

File: loguru/_locks_machinery.py (hunk 2 above)
Comment (jacksmith15) [reply loguru/_locks_machinery.py:44]: 👍

File: loguru/_handler.py (hunk 1 above)
Comment (jacksmith15) [reply loguru/_handler.py:178]: 👍

File: loguru/_handler.py [hunk 6]
```diff
@@ -297,6 +317,7 @@ def _queued_writer(self):
     def __getstate__(self):
         state = self.__dict__.copy()
         state["_lock"] = None
+        state["_lock_acquired"] = None
```
Delgan, line 320: Very good catch.

loguru/_handler.py
```diff
@@ -95,6 +98,24 @@ def __init__(
...
+        """Acquire the lock, but fail fast if its already acquired by the current thread."""
+        if getattr(self._lock_acquired, "acquired", False):
+            raise RuntimeError(
+                "Could not acquire internal lock because it was already in use (deadlock avoided). "
+                "This likely happened because the logger was re-used inside a sink, a signal "
+                "handler or a '__del__' method. This is not permitted because the logger and its "
+                "handlers are not re-entrant."
+            )
+        self._lock_acquired.acquired = True
+        try:
```
Delgan, line 112: I would replace this block with:

```python
self._lock_acquired.acquired = True
with self._lock:
    yield
self._lock_acquired.acquired = False
```

Is theer any reason to use `try / finally`?

tests/test_locks.py
Delgan: Given the unpredictable nature of the garbage collector, I would indeed prefer to call `gc.collect()` to guarantee the test behaves as expected regardless of `gc` configuration and possible change of the `gc` implementation in the future.

It's a good idea to give a concrete example, though. I would suggest to leave it as a comment and/or link to the related ticket you created.

loguru/_handler.py
jacksmith15: I think so. If an exception is raised, then the lock would be released, but the `selc._lock_acquired.acquired` would still be `True`.

Try this fake example:

```python
from contextlib import contextmanager

@contextmanager
def fake_lock():
    try:
        print("lock acquired")
        yield
    finally:
        print("lock release")


@contextmanager
def fake_protected_lock():
    print("lock state set to True")
    with fake_lock():
        yield
    print("lock state set to False")


with fake_protected_lock():
    raise ValueError

```

loguru/_handler.py
jacksmith15: It was well was covered in tests :muscle:

loguru/_handler.py
Delgan: My bad, I don't know why I assumed it was automatically handled, I'm probably too used to `pytest` fixtures. The `try / finally` is required as explained in the docs I shared myself... :man_facepalming: 

Thanks for the example. :+1:

Delgan:
Everything looks fine, PR merged! :tada: 

Thanks a lot for your time and your work. This is a problem which exists since the beginning of Loguru and to which I could not think of an acceptable solution. I'm sure this new explicit `RuntimeError` will be of tremendous help to a lot of users which faced tricky deadlocks.

Issue #209, opened 2020-02-03
No correspondence between string and int level format


Vedrillan:
The behaviors when logging with the level set as "INFO" or 20 is different, though they share the same level number.

![image](https://user-images.githubusercontent.com/2758851/73668271-ef815480-46a5-11ea-84fa-47faf16da1d2.png)

I noticed that because I wanted to use logging levels constant:
```
from logging import INFO
from loguru import logger
logger.log(INFO, "toto")
```

If the level number is registered in loguru it should be matched and used instead of displaying the generic `"Level %d" % level_no` message without any colorization.

It would also be nice to have Loguru providing the level numbers as constant (e.g loguru.INFO), otherwise levels are just magic strings that cannot be easily looked up or referenced.

Delgan:
Hi @Vedrillan.

The way the levels work is indeed slightly different between `logging` and `loguru`. Using `loguru, the he severity level is not mapped against any level name, bu this is made on purpose.

A level is basically a `namedtuple`:
```python
Level(name='INFO', no=20, color='<bold>', icon='ℹ️')
```
I think what should be used to identify a level is its **name**, not its **number**. It's perfectly valid to create a custom level while another one exist with the same `no`, this shall not create any conflict (which would be the case if the `no` was also mapped to the level).

Adding constants to reference levels has been discussed, but it doesn't seem much useful to having to type `loguru.INFO` instead of `"INFO"`. I don't think of them as being magic strings, it's just a level name and could be any other string. Moreover, the levels are not truly constants because they can actually be modified.

Vedrillan:
After more reading of the code, I indeed see that the levels are identified by their names.

This is a strange thing to me, as in my opinion log levels should be identified by both their names and levels. Changing the level or name of a "core" level, such as "INFO", would not make sens as it goes against common logging practices (of things I have used, ruby, php, javascript, syslog). Such changes would just generate confusion.

One could consider that developers are free to make "DEBUG" logs appear more critical that "EMERGENCY" but others prefer to capitalize on standards and common practices. I am obviously of second group, and if you are not that is alright, we all have our opinions and they do not have to be similar.

In the end I am only hoping for the existing levels to be constant, by their names and levels, across all my projects, to avoid the mess and pain that changing the "standardized level" would generate, but if it is not the philosophy of loguru then be it, I will just have to use something else that fits my requirements.

Delgan:
[...]

PR #743, opened 2022-11-05
use static_level_no in explicit methods

tds333:
Small performance improvement if no logging is needed.
Also use the default static_level_no in known methods. Defaults can be used because changes are not allowed.

### DIFF
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -1990,31 +1990,51 @@ class Logger:
 
     def trace(__self, __message, *args, **kwargs):  # noqa: N805
         r"""Log ``message.format(*args, **kwargs)`` with severity ``'TRACE'``."""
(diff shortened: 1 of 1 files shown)


Delgan:
Thanks for the improvement, @tds333!
Level severity can't be changed since #209, indeed, and I didn't realize it meant we could also early exit the function without the (somewhat expensive) `getframe()` call.

Actually, your PR made me realize I could also generalize this improvement to the `.log()` function and also remove the `static_level_no` argument for simplification. See https://github.com/Delgan/loguru/commit/a37d2d12211cfacd2b10d5120ddb2d8d4efea9f6.
It implies an additional `dict` lookup (error-free) compared to your solution, but I think it's totally negligible (and probably faster for the case without early exit).

tds333:
I had the same in mind. But then decided to submit only a minimal change.
I think the refactoring improves and simplifies the code. Thanks

PR #750, opened 2022-11-16
Add note about creating handlers in a different multiprocessing context

TCatshoek:
I'm not sure if this is the right level of detail, or if the examples are necessary. I'd be happy to change it if you want!


Delgan:
That looks great, thanks a lot @TCatshoek for taking the time to write these detailed explanations!

TCatshoek:
Glad I could help! :)

PR #767, opened 2022-12-25
fix: catch KeyError & ValueError when formatting failed

GreyElaina:


GreyElaina:
as designed...

crizCraig:
Why was this closed btw? Running into the same issue... #1008

GreyElaina:
> Why was this closed btw? Running into the same issue... #1008

it donot fix this issue corrently.

odevtmp:
pathetic. this library is not usable. how one is supposed to be sure beforehand that there will be no curlies in log messages? not everything depends on oneself

PR #807, opened 2023-02-27
Fix escaped characters representation in logs.

A-Tarski:
image
Hi! Log entries with non utf-8 characters are printed escaped. I do not see any way to make them be printed in original way. In this pull request is an example how to make logs correct. Not sure that this is a general solution, but maybe will be good starting point to look at.

### DIFF
--- a/loguru/_handler.py
+++ b/loguru/_handler.py
@@ -187,6 +187,8 @@ class Handler:
 
             str_record = Message(formatted)
             str_record.record = record
+            str_record = str_record.encode('ascii').decode('unicode-escape').encode('utf-16', 'surrogatepass').decode(
+                'utf-16')
 
             with self._protected_lock():
                 if self._stopped:


Delgan:
Hi, do you have a reproducible example of the encoding problem you're facing?

Loguru plays well with unicode characters, and Python strings are unicode. Is it a problem related to the encoding of your sink not supporting utf8?

A-Tarski:
> Hi, do you have a reproducible example of the encoding problem you're facing?
> 
> Loguru plays well with unicode characters, and Python strings are unicode. Is it a problem related to the encoding of your sink not supporting utf8?

Hm, the problem is in formatter:

`def _format_local_log_msg(record: dict):
    format_string = LOGURU_FORMAT
    format_string = format_string.replace("{time:YYYY-MM-DD HH:mm:ss.SSS}", "{time:MMM-DD HH:mm:ss}")

    if record["extra"]:
        record["formatted_extra"] = highlight(
            json.dumps(record["extra"], default=str),
            lexer=JsonLexer(),
            formatter=Terminal256Formatter(style="native"),
        )

        format_string += " {formatted_extra}"
        format_string += "{exception}"
    else:
        format_string += "{exception}\n"

    return format_string
    
    local_log_handler = {
        "sink": sys.stdout,
        "level": DEBUG,
        "format": _format_local_log_msg,
        "diagnose": True,
    }

    loguru.logger.configure(
        handlers=[local_log_handler],
    )
    `
    
    On this step symbols are decoded as escaped characters and loguru just print them on this way

Delgan:
The encoded characters you're observing are due to the call to [`json.dumps()`](https://docs.python.org/3/library/json.html#json.dumps). There is default parameter `ensure_ascii` that you should set to `False`.

```python
json.dumps(record["extra"], default=str, ensure_ascii=False),
```

A-Tarski:
That is true, thanks. For few years was sure the problem is somewhere in loguru lib🤦‍♂️
Sorry for disturbing, this fix is useless then.

PR #858, opened 2023-04-29
added telegram handler

vatche-t:
Dear author,

I came across your code on GitHub and was impressed with your implementation of a logger for catching exceptions. I particularly liked the use of logger.catch which makes it easy to catch exceptions and handle them appropriately.

However, I would like to extend the functionality of your logger by adding the ability to send the output to Telegram as well. This would be very useful for me as I often need to receive notifications on my phone when an exception occurs.

I noticed that you have used the logging.StreamHandler class to handle the output of the logger. Would it be possible to modify this to send the output to Telegram as well? Perhaps by creating a custom handler that sends the output to both the stream and Telegram.

I have included a link to my modified code in the attached gist. It would be great if you could take a look and let me know what you think.

Thank you for your great work on this project!

Best regards,
Vatche Thorossian

### DIFF
--- a/loguru/_handler.py
+++ b/loguru/_handler.py
@@ -8,6 +8,7 @@ from threading import Thread
 from ._colorizer import Colorizer
 from ._locks_machinery import create_handler_lock
 
+import logging 
 
 def prepare_colored_format(format_, ansi_level):
     colored = Colorizer.prepare_format(format_)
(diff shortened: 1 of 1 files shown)


Delgan:
Hi @vatche-t, thanks for your interest in contributing.

Actually, Loguru was designed in such a way that it's compatible with pretty much any kind of handler. It does not need to be extended internally to support a Telegram endpoint. What you need is simply to call [`logger.add()`](https://loguru.readthedocs.io/en/stable/api/logger.html#loguru._logger.Logger.add) with an appropriate sink.

```python
class TelegramHandler(logging.Handler):
    def __init__(self, bot, chat_id):
        super().__init__()
        self.bot = bot
        self.chat_id = chat_id

    def emit(self, record):
        log_entry = self.format(record)
        self.bot.send_message(chat_id=self.chat_id, text=log_entry)

handler = TelegramHandler()
logger.add(handler)
```

vatche-t:
thank you very much

PR #883, opened 2023-06-01
improve library disable readme

trim21:
Original example is a little bit confusing, is take me a while to figure out that I need to disable library's `__name__` instead of name I used in `logger.bind(name='my_library')`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -377,8 +377,6 @@ Using the logger in your scripts is easy, and you can |configure|_ it at start. 
 
     # For libraries, should be your library's `__name__`
     logger.disable("my_library")
-    # or
-    logger.disable(__name__)
 
     logger.info("No matter added sinks, this message is not displayed")
     logger.enable("my_library")
--- a/README.rst
+++ b/README.rst
@@ -375,7 +375,7 @@ Using the logger in your scripts is easy, and you can |configure|_ it at start. 
     }
     logger.configure(**config)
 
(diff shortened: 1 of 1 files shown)


Delgan:
Hi @trim21, thanks for the improvement suggestion.

Could you please clarify your use case?
I have trouble understand why you would need to use `__name__`.

What I mean by "library" is some module intended to be used as an external source (after being installed). If the module is part of a project, it does not need to call `logger.disable()`, it can be configured at the entry point of the project instead.

trim21:
> Hi @trim21, thanks for the improvement suggestion.
> 
> Could you please clarify your use case? I have trouble understand why you would need to use `__name__`.

I'm using loguru in my library and I find that I didn't actually disable it because I'm using the wrong name.

For your example, `pip install my-library`, and in the code, I should call `logger.disable('my_library')` instead of `my-library`, or `logger.disable('my-library-2')` for `logger.bind(name='my-library-2')`

Delgan:
Ok, I see, thanks.

I think suggesting `logger.disable(__name__)` can be confusing as well, because the result will depend on where it's called from (while using a static name like `"my_library"` is less error-prone in my opinion).

Can you please change your commit so that the only change is:

```python
logger.disable("my_library")  # Should be your library's `__name__` 
```

I think it helps clarify usage, right?

Delgan:
Thank you!

Issue #342, opened 2020-10-01
ctypes pickling error


JulianOrteil:
Good day, Delgan.

I have an error I am encountering in the handlers of loguru. This error is being raised when I use "logger.exception(msg)" in the except block of a try statement.

**Some context:**
I'm developing an application to interface with NVIDIA's AI Triton server. The server's client-side python bindings are written in both C/C++ and Python. I'm mentioning this because I am developing on Windows, and replicating this issue is fairly hard due to specifics in the environment for the Triton server. I can give you the files and procedures to set up an environment like mine for this if you'd like, but I'll leave that up to you.

**The error:**
*Stacktrace:*
```
Traceback (most recent call last):
  File "C:\Users\jules\.conda\envs\app\lib\site-packages\loguru\_handler.py", line 175, in emit
    self._queue.put(str_record)
  File "C:\Users\jules\.conda\envs\app\lib\multiprocessing\queues.py", line 362, in put
    obj = _ForkingPickler.dumps(obj)
  File "C:\Users\jules\.conda\envs\app\lib\multiprocessing\reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
ValueError: ctypes objects containing pointers cannot be pickled
```
*Code throwing the issue:*
```
try:
    # <Code here raises 'InferenceServerException'>
except InferenceServerException: # The specific exception being raised (Triton server-specific)
    logger.exception("An issue occurred while contacting the inference server:\n")
```
The exception being raised does have calls to dll's in it, so that could possibly be the root cause? Again, it is difficult to reproduce this error because of the environment.

**Exception data as returned by 'sys.exc_info()':**
```
type: <class 'tensorrtserver.api.InferenceServerException'>
value: InferenceServerException(c_void_p(2735452554832)) # This is what I think is causing the issue
traceback: <normal traceback>
```

*Logger setup code:*
```
import os
import sys

import stackprinter
from loguru import logger


class FilterLevel(object):
    """Used to filter logs based dynamic logging level by using the
    :prop:`level` property which can be either an :obj:`int` or
    :obj:`str`.
    """

    _level: Union[str, int]

    def __init__(self) -> None:
        super().__init__()

        self.level = "INFO"

    def __call__(self, record: dict) -> bool:
        levelno = logger.level(self.level).no
        return record["level"].no >= levelno

    @property
    def level(self) -> Union[str, int]:
        return self._level

    @level.setter
    def level(self, value: Union[str, int]) -> None:

[...]

PR #905, opened 2023-06-28
_recattrs.py: take into account exception coming from other classes

ncoudene:
Hello and thank you for this great module !

Recently I have beginning to use loguru with an 'aiohttp' project and I have beginning to encounter some annoying exception coming from loguru :

```
[...]
project  |     raise ClientResponseError(
project  | aiohttp.client_exceptions.ClientResponseError: 400, message='Expected HTTP/', url=URL('https://xxxxxxxxx')
project  | 
project  | During handling of the above exception, another exception occurred:
project  | 
project  | Traceback (most recent call last):
project  |   File "/usr/local/lib/python3.11/site-packages/loguru/_handler.py", line 195, in emit
project  |     self._queue.put(str_record)
project  |   File "/usr/local/lib/python3.11/multiprocessing/queues.py", line 371, in put
project  |     obj = _ForkingPickler.dumps(obj)
project  |           ^^^^^^^^^^^^^^^^^^^^^^^^^^
project  |   File "/usr/local/lib/python3.11/multiprocessing/reduction.py", line 51, in dumps
project  |     cls(buf, protocol).dump(obj)
project  |   File "/usr/local/lib/python3.11/site-packages/loguru/_recattrs.py", line 73, in __reduce__
project  |     pickle.dumps(self.value)
project  | TypeError: can't pickle multidict._multidict.CIMultiDictProxy objects
project  | --- End of logging error ---

```

With the proposed patch it seems to work flawlessly again.

Let me know if it seems fine to you too.

### DIFF
--- a/tests/test_add_option_enqueue.py
+++ b/tests/test_add_option_enqueue.py
@@ -12,12 +12,7 @@ from .conftest import default_threading_excepthook
 
 class NotPicklable:
     def __getstate__(self):
-        raise Exception(
-            [
-                TypeError("You shall not serialize me!"),
-                pickle.PicklingError("You shall not serialize me!"),
-            ],
-        )
+        raise pickle.PicklingError("You shall not serialize me!")
 
     def __setstate__(self, state):
         pass
@@ -90,7 +85,7 @@ def test_caught_exception_queue_put(writer, capsys):
     assert out == ""
     assert lines[0] == "--- Logging error in Loguru Handler #0 ---"
     assert re.match(r"Record was: \{.*Bye bye.*\}", lines[1])
-    assert len(re.findall(r"You shall not serialize me!", err)) == 2
+    assert lines[-2].endswith("PicklingError: You shall not serialize me!")
     assert lines[-1] == "--- End of logging error ---"
 
 
@@ -134,7 +129,7 @@ def test_not_caught_exception_queue_put(writer, capsys):
 
     logger.info("It's fine")
 
-    with pytest.raises(Exception, match=r"You shall not serialize me!"):
+    with pytest.raises(pickle.PicklingError, match=r"You shall not serialize me!"):
         logger.bind(broken=NotPicklable()).info("Bye bye...")
 
     logger.remove()
--- a/tests/test_add_option_enqueue.py
+++ b/tests/test_add_option_enqueue.py
@@ -18,6 +18,14 @@ class NotPicklable:
         pass
 
 
+class NotPicklableTypeError:
+    def __getstate__(self):
+        raise TypeError("You shall not serialize me!")
+
+    def __setstate__(self, state):
+        pass
+
+
 class NotUnpicklable:
     def __getstate__(self):
         return "..."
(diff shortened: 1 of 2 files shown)


Delgan:
Hi @ncoudene, thanks for the improvement!

  You're right that `PickleError` isn't the only exception that can be thrown., `TypeError` and a whole bunch of other errors can occur when `pickle.dumps()` is called. I think we should actually make `try / except` even more generic. This would make it possible to fix this issue: https://github.com/Delgan/loguru/issues/342

Would you mind updating the PR to catch `Exception` instead of `TypeError` and `PickleError` please?

Also, implementing an unit test would be a nice addition. You can copy the test here but replace or extend `NotPickable()` so that it throws `TypeError` instead of `PicklingError`: https://github.com/Delgan/loguru/blob/d379fedd9b19401e937c63de0154caf9a11252ec/tests/test_add_option_enqueue.py#L235-L254

ncoudene:
Hi @Delgan,

I updated the PR with your requests :

- Exception is `Exception` instead of `TypeError` and `PickleError`
- `NotPickable()` throws `Exception` with a list of Exception and the tests have been fixed to catch it accordingly 

Unfortunately the push to codecov seems to have failed.

tests/test_add_option_enqueue.py
```diff
@@ -12,7 +12,12 @@
 
 class NotPicklable:
     def __getstate__(self):
-        raise pickle.PicklingError("You shall not serialize me!")
```
Delgan, line 15: Thanks for the update, however this is not exactly what I had in mind, sorry for the confusion.

With this change, we simulate a generic `Exception` occurring during `pickle.dumps()`. The `TypeError()` and `pickle.PicklingError()` in the list are just arguments of the `Exception` and are not actually very relevant to the test.

I actually expected two distinct test cases:
* one with `NotPicklable` causing a `pickle.PicklingError()` (the common case) ;
* another test with `NotPicklable` causing a `TypeError()` (the improvement you suggested).

In this way, we cover different possible cases of error that can happen.

However, I understand that you might not be very familiar with the code base. I would suggest that you simply duplicate the original `NotPicklable` error and create a new `NotPicklableTypeError` class (which raises a `TypeError` instead of `pickle.PicklingError`). Then you could also duplicate `test_caught_exception_queue_put()` and `test_not_caught_exception_queue_put()` to test against this new `NotPicklableTypeError()` class.

Duplication is fine, I'll find a way to factorize it eventually, but we can keep it simple for now.

Delgan:
Hi @ncoudene.

Thanks for taking the time to update the PR, I've left a comment to clarify how I think testing can be improved. If you don't have the time, let me know, I can always update your branch on my end.

Regarding the `codecov` error, I made an update the other day. I think if you `rebase` your branch against `master`, it should work.

ncoudene:
Hi @Delgan ,

I revert my change and add the change you asked.

Codecov (or github actions) seems buggy.

Let me know if you have some comments.

Delgan:
Great, thanks for taking the time to update the PR. Looks perfect. :+1: 

However, I think something went wrong during your `rebase` because some commits from the `loguru/master` branch were duplicated in your PR. Note that I therefore modified and force-pushed on your branch to remove them.

PR #943, opened 2023-08-13
Fix STRIKE and HIDE ANSI codes

tunaflsh:
The ANSI codes for STRIKE and HIDE are incorrect. According to [ANSI Escape Codes](https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797#colors--graphics-mode) [`STRIKE`](https://github.com/Delgan/loguru/blob/d17ae011117e46365f3b68ec4eded733b8434090/loguru/_colorizer.py#L13) should be `9` and [`HIDE`](https://github.com/Delgan/loguru/blob/d17ae011117e46365f3b68ec4eded733b8434090/loguru/_colorizer.py#L14) should be `8`.

### DIFF
--- a/loguru/_colorizer.py
+++ b/loguru/_colorizer.py
@@ -10,8 +10,8 @@ class Style:
     UNDERLINE = 4
     BLINK = 5
     REVERSE = 7
-    STRIKE = 8
-    HIDE = 9
+    HIDE = 8
+    STRIKE = 9
     NORMAL = 22
 
 


Delgan:
Good catch, thanks!

Can you please remove the commits related to the other PR so I can squash-merge this one?

loguru/_colorizer.py
```diff
@@ -10,8 +10,8 @@ class Style:
     UNDERLINE = 4
     BLINK = 5
     REVERSE = 7
-    STRIKE = 8
-    HIDE = 9
+    STRIKE = 9
```
Delgan, line 13: Can you please swap the order of `HIDE` and `STRIKE` so that the list stays sorted?

tunaflsh: Sure, I'll do that within a few days.

Delgan:
I rebased your branch before the merge.

Thanks for the fix @tunaflsh!

PR #953, opened 2023-08-26
update example to use safe permissions

quantumpacket:
- Remove unnecessary umask modification
- Change permission from world read/write/execute to owner read/write

### DIFF
--- a/docs/resources/recipes.rst
+++ b/docs/resources/recipes.rst
@@ -14,6 +14,7 @@ Code snippets and recipes for ``loguru``
 .. |contextlib.redirect_stdout| replace:: :func:`contextlib.redirect_stdout`
 .. |copy.deepcopy| replace:: :func:`copy.deepcopy`
 .. |os.fork| replace:: :func:`os.fork`
+.. |os.umask| replace:: :func:`os.umask`
 .. |multiprocessing| replace:: :mod:`multiprocessing`
 .. |pickle| replace:: :mod:`pickle`
 .. |traceback| replace:: :mod:`traceback`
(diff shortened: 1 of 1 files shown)


Delgan:
Thanks!

However I think it's better to keep some reference to the `os.umask(0)` workaround. As you said in https://github.com/Delgan/loguru/commit/9aee1fbfc4c928828bc0791503c35fe4fc57350a#r125574708, snippets are often copy-pasted without much thoughts and I expect that some user will not `os.opener()` just mask the existing permissions.
I believe it's worth mentioning why and when calling [`os.umask()`]() is required, while emphasizing safety considerations.

quantumpacket:
Correct me if I'm wrong, but calling `os.umask(0)` without using the opener will default the file permissions to `666`, which would make the file read/write for any user.

```python
os.umask(0)

logger.add(
    sink='loguru.log',  # 666 (unsafe as world read/write)
    # ...
)
```


If the user wants to set the file permissions using umask without using `os.opener()` then they should be using `os.umask(0o177)` to make it `600`.

```python
os.umask(0o177)

logger.add(
    sink='loguru.log',  # 600 (read/write by owner only)
    # ...
)
```

So using only the above umask would work, or using only the opener, but no need for both.

However, as far as I know using `os.umask()` would make it apply to all Python code that are creating files/directories and cause unintended side-effects if not handled carefully. You would usually get the return value (previous umask) and then set the umask back to the original once done with whatever file/directory operations.

Delgan:
You're right, thanks for pointing that out. `os.umask(0)` is not safe, it was a easy workaround but we should not suggest it.  I think we should instead quickly explain to the user how file permissions work in regards to the default `umask`. Calling `os.umask()` should be used as last resort, but it's a requirement if user wants `666` by default for example.

I suggest to add a paragraph at the very end of the section to explain how it works. Something along those lines:

> Note that the provided mode will be masked out by the OS [umask](https://en.wikipedia.org/wiki/Umask) value (which describes which bits are _not_ to be set when creating a file or directory). This value is conventionally equals to `0o022`, which means specifying a `0o666` mode will result in a `0o666 - 0o022 = 0o644` file permission in this case (which is actually the default). It is possible to change the umask value by first calling [`os.umask()`](https://docs.python.org/3/library/os.html#os.umask), but this needs to be done with careful consideration, as it changes the value globally and can cause security issues.

Delgan:
@quantumpacket Thoughts on the suggested paragraph? I can update the PR and merge it if you want.

quantumpacket:
@Delgan Looks good to me. If you can update the PR please, I'm terrible at using git.

Delgan:
No problem, PR updated and merged, thanks for the improvement. :+1:

PR #955, opened 2023-08-26
Replace flake8 and isort with ruff

### LINKED ISSUE #830: [Enhancement] Use ruff instead of flake8 for linting in pre-commit
I guess using ruff instead of flake8 could be considered, especially for performance reasons of lining + additional features like import sorting (like isort) and removing unused code (like autoflake).

See here: https://github.com/charliermarsh/ruff

(There is also a performance comparison between common linters)

hillairet:
Fix #830 

I took on the little task of switching to ruff.
It seems to be working properly but I'm seeing a lot of [N805](https://beta.ruff.rs/docs/rules/invalid-first-argument-name-for-method/), and [B018](https://beta.ruff.rs/docs/rules/useless-expression/) errors.
There are also a handful of other B*** errors.

I'm confused by that since ruff should be reimplementing the same errors as bugbear and pep8-naming. Somehow though, ruff seems more stringent. But maybe I missed some excludes from flake8?
I did introduce a bugbear error and checked on the master branch that it got picked up by flake8 so it is running.

@Delgan What would be your preferred solution? Do you know why this might be happening?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,20 +12,13 @@ repos:
     args: [--autofix]
   - id: pretty-format-yaml
     args: [--autofix, --indent, '2']
-- repo: https://github.com/pycqa/isort
-  rev: 5.12.0
-  hooks:
-  - id: isort
 - repo: https://github.com/ambv/black
   rev: 23.7.0
   hooks:
   - id: black
     args: [-l, '100', --target-version, py35]
-- repo: https://github.com/pycqa/flake8
-  rev: 6.1.0
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  rev: v0.0.286
   hooks:
-  - id: flake8
-    args: [--exclude, tests/exceptions/source]
-    additional_dependencies:
-    - flake8-bugbear==23.1.20
-    - pep8-naming==0.13.3
+  - id: ruff
+    args: [--fix, --exit-non-zero-on-fix]
--- a/ruff.toml
+++ b/ruff.toml
@@ -1,0 +1,4 @@
+# Enforce isort (I), bugbears (B), and pep8-naming (N) rules
+select = ["E", "F", "I", "B", "N"]
+line-length = 100
+exclude = ["tests/exceptions/source"]
--- a/ruff.toml
+++ b/ruff.toml
@@ -1,6 +1,8 @@
 # Enforce isort (I), bugbears (B), and pep8-naming (N) rules
-select = ["E", "F", "I", "B", "N"]
+select = ["E", "F", "W", "I", "B", "N"]
 # Ignore these errors for now that are picked up by ruff but were not by flake8
 ignore = ["B904", "B033", "B028", "B018"]
 line-length = 100
 exclude = ["tests/exceptions/source", "loguru/__init__.pyi"]
+[pycodestyle]
+max-doc-length = 100
--- a/ruff.toml
+++ b/ruff.toml
@@ -1,4 +1,6 @@
 # Enforce isort (I), bugbears (B), and pep8-naming (N) rules
 select = ["E", "F", "I", "B", "N"]
+# Ignore these errors for now that are picked up by ruff but were not by flake8
+ignore = ["B904", "B033", "B028", "B018"]
 line-length = 100
-exclude = ["tests/exceptions/source"]
+exclude = ["tests/exceptions/source", "loguru/__init__.pyi"]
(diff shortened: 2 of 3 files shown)


hillairet:
Ok I understand things more now:
1. flake8 doesn't check `*.pyi` files it seems. Ruff does by default. So I added an exclusion for that
2. flake8-bugbear was on version 23.1.20 which had [errors up to B027](https://github.com/PyCQA/flake8-bugbear/tree/23.1.20#list-of-warnings) so I excluded B033 and B028
3. Flake8-bugbear errors B9xx are opinionated and therefore not enabled by default, but they are in ruff. So I excluded B904.
4. For some reason flake8-bugbear 23.1.20 doesn't pick up `1 / 0` as a useless expression or B018, but ruff does so I excluded that too.

I'm happy to add some `noqa: B***` instead if you prefer silencing individual lines.
Also something like [B904](https://beta.ruff.rs/docs/rules/raise-without-from-inside-except/) just requires to use `raise ... from ValueError` so it's easy to fix.
Let me know what you prefer.

Delgan:
That is really great, thanks for working on this migration! :tada: 

I would like to remove the global `ignore` and fix the reported error or add `noqa` comment on a case-per-case basis, and also avoid excluding `"loguru/__init__.pyi"`.
However, we can do this in subsequent pull requests. It's a good idea to have `ruff` introduced this way and preserve backward-compatibility as a first step.

A few remarks regarding this PR, though:
* Can we now remove [the `flake8` configuration](https://github.com/Delgan/loguru/blob/3876de0ca7b8a80f9359149be3dd7686fe192d9a/tox.ini#L30-L39) from the `tox.ini` file?
* What about [the `W` rules from `pycodestyle`](https://beta.ruff.rs/docs/rules/#warning-w) and the `max-docs-length` configuration?
* Am I understanding correctly that `ruff` supersedes `isort`? If so, shouldn't we migrate [the `isort` configuration](https://github.com/Delgan/loguru/blob/3876de0ca7b8a80f9359149be3dd7686fe192d9a/tox.ini#L26-L28) from the `tox.ini` file?
* There are several remaining `flake8: noqa` instruction [in some files](https://github.com/Delgan/loguru/blob/3876de0ca7b8a80f9359149be3dd7686fe192d9a/tests/exceptions/source/diagnose/attributes.py#L2), can we convert them to `ruff`?

hillairet:
I am also a big fan of step by step approach so going forward with the global ignores and going line by line later sounds great. I'll be happy to work on the next PR too. :+1: 

About the remarks:
1. Oh yeah the tox.ini! I missed that. :sweat_smile: I'll take care of it.
2. I can add them and see what happens. I just need to add `W` to the `select` list in ruff.toml
3. Yes ruff implements isort. It's the `I` in ruff.toml. I'll do that too.
4. All of the `flake8: noqa` are in files under `tests/exceptions/source` which was excluded for flake8 in .pre-commit-config.yaml ... so I excluded this folder from ruff's processing as well. Do you need these `noqa` for editor purposes? In any case I can replace `flake8` with `ruff`. It's the same syntax.

Delgan:
Regarding the `#flake8: noqa` comment, I don't remember exactly why I added them in https://github.com/Delgan/loguru/commit/7ef57c3e6d8ad7dc2bcebf6d32465d306d3f0acb. Currently, removing them causes `flake8` to fail during `pre-commit`, but replacing `args: [--exclude, tests/exceptions/source]` with `args: [--exclude, tests/exceptions/source/*]` in `.pre-commit-config.yaml` solves the issue. Maybe I didn't realize this workaround at that time, or maybe it wasn't available. 

Anyway, I suppose we should be able to remove the `#flake8: noqa` comments, but we can do this in a future PR, as it implies updating the generated `tests/exceptions/output` files.

hillairet:
@Delgan So I did the flake8 and isort clean up in tox.ini and added the pycodestyle warnings.
I set the `max-doc-length` to 100 because it was effectively enforced before by the `line-length`. I checked that at 90 it finds a bunch of issues.

And to summarize on the future work:
1. Remove the global error ignores `"B904", "B033", "B028", "B018"`
2. Remove the ignore of `loguru/__init__.pyi`
3. Remove the `#flake8: noqa` from the files in `tests/exceptions/source` and use an exclude `tests/exceptions/source/*` instead.

I'll create issues for the these :point_up: and I'm happy to take care of them.

Delgan:
Great work! Thanks for the thoughtful updates. :+1:

Issue #954, opened 2023-08-26
Disable LOGURU_DIAGNOSE by default - credential leak risk


aedwardstx:
In the interest of being secure by default, `LOGURU_DIAGNOSE`(better_exceptions) should be disabled by default due to the potential for leaking credentials in logs. Per the guidance in https://github.com/Qix-/better-exceptions/issues/119#issuecomment-1024885652 better_exceptions should not be used when working with sensitive data. Per https://github.com/Delgan/loguru/blob/master/README.rst?plain=1#L242 "Caution, may leak sensitive data in prod". The way the readme describes enabling the feature and highlights the risk, it would be reasonable to assume that its disabled by default but this is not the case. 

https://github.com/Delgan/loguru/blob/d17ae011117e46365f3b68ec4eded733b8434090/loguru/_defaults.py#L43

Delgan:
Thank you for your input regarding the default value of `LOGURU_DIAGNOSE`.

Balancing between ease of use and security is a crucial consideration. While the display of local variables is indeed a feature that raises security concerns, it's also a highly regarded aspect of Loguru. The library is designed with the intention of offering a seamless and user-friendly experience from the moment it's used.

The default settings in Loguru are primarily tailored for swift development iterations. However, as the software matures, it becomes imperative for developers to delve into additional aspects, such as the one you've pointed out. The documentation covers these concerns comprehensively under the section [Security considerations when using Loguru](https://loguru.readthedocs.io/en/stable/resources/recipes.html#security-considerations-when-using-loguru).

In essence, while I acknowledge the significance of enhancing security, transforming Loguru into a completely secure-by-default tool is not possible without undermining its default usability. As a result, I'm inclined to provide users with the responsibility to configure the logger to meet their specific security requirements when transitioning to production environments.

aedwardstx:
Thank you for your response and for developing such an intuitive library! I understand your stance and it seems reasonable. Per my comment about clarity in the readme, I've submitted https://github.com/Delgan/loguru/pull/957 for your consideration.

Delgan:
Glad we've found a compromise. :smile: 

Thanks for your contribution.

PR #957, opened 2023-08-27
Add note about diagnose=True being default

aedwardstx:
Per #954,

Addresses clarity in the readme around `diagnose=True` being the default and provides a link to the security considerations doc.


Delgan:
Great, thank you. :+1:

Issue #968, opened 2023-09-02
Resolve the new BXXX that appeared when switching to ruff


hillairet:
As discussed in [this PR](https://github.com/Delgan/loguru/pull/955#issuecomment-1694694943) a few linting and validation errors appeared when we switched from flake8 to ruff.
To get the PR merged, global exclusions of the errors were added to the ruff configuration.
Instead we should deal with each error one by one by fixing the issue or adding a `noqa` wherever it's justified to do so.
For example some (all?) B018 complains about `1 / 0` being a useless expression. But it's part of a test checking that the exception is handled correctly so `noqa` is justified.

hillairet:
@Delgan Question about the errors in `loguru/__init__.pyi`.
Why did you use `__self` rather than `self` for some methods?
You ignored this error in `_logger.py` so I'll ignore line by line in `loguru/__init__.pyi` as well. The question is because I'm curious.  :wink:

Delgan:
Hi @hillairet. :)

I use `__self` in `loguru/__init__.pyi` to match the signature of the `logger` methods defined in `_logger.py`.  These class methods are implemented using `__self` instead of the typical `self` to mitigate the risk of conflicts arising from arbitrary keys employed in `**kwargs` arguments.

In other words, it makes it possible to do the following:

```python
self = "delgan"
logger.info("I'm {self}", self=self)
```

Otherwise, this would raise an exception:

```
TypeError: Logger.info() got multiple values for argument 'self'
```

This also applies to the `__message` parameter.

Ideally, we could use [PEP 570 – Python Positional-Only Parameters](https://peps.python.org/pep-0570/) which is designed for such cases. However, this is only available for Python 3.8, and Loguru supports older versions of Python that are not compatible with this new syntax. Therefore, I prefixed the parameters with "__" as it's a convention used by some libraries to signal their are positional-only.

PR #970, opened 2023-09-02
Remove ruff config excludes

hillairet:
Fix: #968

So I removed all the ignores of specific errors and the exclude of `loguru/__init__.pyi`

* For `loguru/__init__.pyi` I just added `noqa: N805` (thanks for the detailed explanation @Delgan )
* For [B018](https://beta.ruff.rs/docs/rules/useless-expression/), it was on a bunch of `1 / 0` in the tests so I added `noqa: B018`
* For [B028](https://beta.ruff.rs/docs/rules/no-explicit-stacklevel/), I made added `stacklevel=1` which is the default value but bugbear wants this to be explicit. So no change of behavior expected.
* For [B033](https://beta.ruff.rs/docs/rules/duplicate-value/), it was simply a duplicated entry in a set in `tests/test_filesink_retention.py` so I removed the duplicate.
* :warning: For [B904](https://beta.ruff.rs/docs/rules/raise-without-from-inside-except/), I added a `from` clause because the code was reraising an exception without it. That one needs your review @Delgan!


loguru/_logger.py
```diff
@@ -890,11 +890,11 @@ def add(
                 elif isinstance(level_, str):
                     try:
                         levelno_ = self.level(level_).no
-                    except ValueError:
+                    except ValueError as err:
                         raise ValueError(
                             "The filter dict contains a module '%s' associated to a level name "
                             "which does not exist: '%s'" % (module, level_)
-                        )
+                        ) from err
```
Delgan, line 897: Can we just `raise` from `None` here please? It's consistent with other raised `Exception` in the module. I prefer to hide implementation details from the stack traceback.

loguru/_logger.py
```diff
@@ -1378,6 +1378,7 @@ def opt(
             warnings.warn(
                 "The 'ansi' parameter is deprecated, please use 'colors' instead",
                 DeprecationWarning,
+                stacklevel=1,
```
Delgan, line 1381: Given than [`B028`](https://beta.ruff.rs/docs/rules/no-explicit-stacklevel/) considers the default `stacklevel=1` as bad, shouldn't we replace it with `stacklevel=2`?

It seems indeed best practice, I didn't realize that the warnings weren't even visible to the user by default...

It slightly changes the behavior but this is for the better in my opinion.

loguru/_logger.py
hillairet: Makes sense. Done.

loguru/_logger.py
hillairet: Sure thing. Done.

Delgan:
This is perfect, I'm really glad to see all these code smells addressed!
Thanks a lot for your work and all the details. :+1:

Issue #654, opened 2022-05-13
The test suite crashes pytest on Python 3.11.0b1


mgorny:
```pytb
$ tox -e py311
GLOB sdist-make: /tmp/loguru/setup.py
py311 create: /tmp/loguru/.tox/py311
py311 inst: /tmp/loguru/.tox/.tmp/package/1/loguru-0.6.0.zip
py311 installed: alabaster==0.7.12,attrs==21.4.0,Babel==2.10.1,black==22.3.0,certifi==2021.10.8,charset-normalizer==2.0.12,click==8.1.3,colorama==0.4.4,coverage==6.3.2,distlib==0.3.4,docutils==0.16,filelock==3.6.0,flake8==4.0.1,freezegun==1.1.0,idna==3.3,imagesize==1.3.0,iniconfig==1.1.1,isort==5.10.1,Jinja2==3.1.2,livereload==2.6.3,loguru==0.6.0,MarkupSafe==2.1.1,mccabe==0.6.1,mypy-extensions==0.4.3,packaging==21.3,pathspec==0.9.0,platformdirs==2.5.2,pluggy==1.0.0,py==1.11.0,pycodestyle==2.8.0,pyflakes==2.4.0,Pygments==2.12.0,pyparsing==3.0.9,pytest==7.1.2,pytest-cov==3.0.0,python-dateutil==2.8.2,pytz==2022.1,requests==2.27.1,six==1.16.0,snowballstemmer==2.2.0,Sphinx==4.5.0,sphinx-autobuild==2021.3.14,sphinx-rtd-theme==1.0.0,sphinxcontrib-applehelp==1.0.2,sphinxcontrib-devhelp==1.0.2,sphinxcontrib-htmlhelp==2.0.0,sphinxcontrib-jsmath==1.0.1,sphinxcontrib-qthelp==1.0.3,sphinxcontrib-serializinghtml==1.1.5,toml==0.10.2,tomli==2.0.1,tornado==6.1,tox==3.25.0,urllib3==1.26.9,virtualenv==20.14.1
py311 run-test-pre: PYTHONHASHSEED='300318784'
py311 run-test: commands[0] | flake8 --exit-zero loguru tests docs setup.py
py311 run-test: commands[1] | pytest --cov loguru/ --cov-report=xml
========================================================= test session starts =========================================================
platform linux -- Python 3.11.0b1, pytest-7.1.2, pluggy-1.0.0
cachedir: .tox/py311/.pytest_cache
[...]

Issue #449, opened 2021-05-24
Mypy complains with type-hints


rafaellehmkuhl:
First of all, this library is really, really handy. Logging is so much easier with it. Thanks!
Regarding the issue:

I'm working on a project that enforces type-hints with Mypy, and I'm intercepting some logging activity from another process, thus using [the InterceptHandler documented on the README file](https://github.com/Delgan/loguru/blob/master/README.rst#entirely-compatible-with-standard-logging):

```python
# main.py

from loguru import logger

class InterceptHandler(logging.Handler):
    def emit(self, record) -> None:
        # Get corresponding Loguru level if it exists
        try:
            level = logger.level(record.levelname).name
        except ValueError:
            level = record.levelno

        # Find caller from where originated the logged message
        frame, depth = logging.currentframe(), 2
        while frame.f_code.co_filename == logging.__file__:
            frame = frame.f_back
            depth += 1

        logger.opt(depth=depth, exception=record.exc_info).log(level, record.getMessage())


logging.basicConfig(handlers=[InterceptHandler()], level=0)

logger.remove()
logger.add(sys.stderr, format="<green>{time:HH:mm:ss.SSS}</green> | <level>{message}</level>")

logger.info("Hi!")
```

This is working perfectly, but Mypy is complaining about the missing type-hints:

```
main.py:4: error: Function is missing a type annotation
main.py:14: error: Incompatible types in assignment (expression has type "Optional[FrameType]", variable has type "FrameType")
Found 2 errors in 1 file (checked 1 source file)
```

The first one is noticing that the `record` argument is not typed and the second one is noticing the type of the variable is changing.

Is there a recommended way of solving them?

I tried to import `Record` from loguru to type-hint the `record` argument but got several errors after:

```
main.py:4: error: Argument 1 of "emit" is incompatible with supertype "Handler"; supertype defines the argument type as "LogRecord"
main.py:4: note: This violates the Liskov substitution principle
main.py:4: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
main.py:7: error: "Record" has no attribute "levelname"
main.py:9: error: "Record" has no attribute "levelno"
main.py:17: error: "Record" has no attribute "exc_info"
main.py:17: error: "Record" has no attribute "getMessage"
Found 5 errors in 1 file (checked 1 source file)
```

Delgan:
Hey @rafaellehmkuhl, gad you like Loguru, thanks!

[...]

PR #971, opened 2023-09-03
Improve typing of PropagateHandler and InterceptHandler

Delgan:
See https://github.com/Delgan/loguru/issues/654#issuecomment-1704027394 and https://github.com/Delgan/loguru/issues/449#issuecomment-1703972428.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -406,7 +406,7 @@ Need to propagate `Loguru` messages to standard `logging`?
 ::
 
     class PropagateHandler(logging.Handler):
-        def emit(self, record):
+        def emit(self, record: logging.LogRecord) -> None:
             logging.getLogger(record.name).handle(record)
 
     logger.add(PropagateHandler(), format="{message}")
@@ -416,16 +416,17 @@ Want to intercept standard `logging` messages toward your `Loguru` sinks?
 ::
 
     class InterceptHandler(logging.Handler):
-        def emit(self, record):
+        def emit(self, record: logging.LogRecord) -> None:
             # Get corresponding Loguru level if it exists.
+            level: str | int
             try:
                 level = logger.level(record.levelname).name
             except ValueError:
                 level = record.levelno
 
             # Find caller from where originated the logged message.
-            frame, depth = sys._getframe(6), 6
-            while frame and frame.f_code.co_filename == logging.__file__:
+            frame, depth = inspect.currentframe(), 0
+            while frame and (depth == 0 or frame.f_code.co_filename == logging.__file__):
                 frame = frame.f_back
                 depth += 1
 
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -416,16 +416,17 @@ Want to intercept standard `logging` messages toward your `Loguru` sinks?
 ::
 
     class InterceptHandler(logging.Handler):
-        def emit(self, record):
+        def emit(self, record: logging.LogRecord) -> None:
             # Get corresponding Loguru level if it exists.
+            level: str | int
```
quantumpacket, line 421: The typehint can actually be put on the first declaration as it's the first declaration that is being used to determine the type for the same named variable. This works for Mypy, Pyright and Pyre.

```python
        try:
            level: int | str = logger.level(record.levelname).name
        except ValueError:
            level = record.levelno
```

Delgan: @quantumpacket Is it considered best practice? I felt it might be confusing for the reader, so I decided to explicitly declare the `level` types before the assignment.

quantumpacket: Not sure what would consider it best practice or not, it's allowed by all type checkers listed. I guess it's a matter of preference at this point. :man_shrugging:

Delgan: Ok, I'll leave it as is then, but thanks for the heads-up.

quantumpacket:
Was up all night reading about frames/stacks and I came up with another way to get the `depth` value, which works in my tests _(need your confirmation)_, and is a bit cleaner with reduced cyclomatic complexity.

```python
depth = 1

for frame in traceback.extract_stack():
    if frame.filename == logging.__file__:
        depth += 1
```
Python, a thousand and one ways to do the same thing. :laughing:

Delgan:
Haha, nice finding! 
It passes all tests in [`tests/test_interception.py`](https://github.com/Delgan/loguru/blob/2c585a1c186b4dadd05faa4c21363c9479e151b4/tests/test_interception.py) so I think you're good!

This implementation indeed looks pretty good, at the cost of a bit more memory usage. I personally prefer to deal directly with the `frame` object than the `FrameSummary` abstraction.

quantumpacket:
@Delgan if you still want to use `inspect.currentframe()` then this should work _(works on my tests)_:

```python
        frame, depth = inspect.currentframe(), 1

        while frame:
            if inspect.getfile(frame) == logging.__file__:
                depth += 1

            frame = frame.f_back
```

Delgan:
I don't really have any particular preference. I've just documented a solution that solves the problems you mentioned, so I didn't dwell on it too much since I'm planning to replace these classes anyway. But thanks for sharing.

Issue #974, opened 2023-09-07
RuntimeError: context has already been set. When I use both Sanic and loguru


ChrisYP:
```python
from sanic import Sanic, text
from loguru import logger

app = Sanic(__name__)


@app.get('/')
async def run_shell(request):
    return text('Hello world!')


if __name__ == '__main__':
    app.run(host='0.0.0.0', port=6666, debug=True, access_log=True, workers=2)
```
python version:3.8
Name: sanic
Version: 23.6.0
Name: loguru
Version: 0.7.1

The above code thorws an error:
    raise RuntimeError('context has already been set')
RuntimeError: context has already been set

Delgan:
Thanks for the report and sorry for the inconvenience.

For `multiprocessing` purposes, Loguru v0.7.1 added a new `context` argument and uses [`multiprocessing.get_context()`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.get_context) by default. This is supposed to have the same effect as using `multiprocessing` directly, but makes it easier to switch from one context to another (eg. from `"fork"` to `"spawn"` on Linux).
Unfortunately, I didn't notice that calling `multiprocessing.get_context()` had the side effect of also **setting** the default context globally... And once a context is set, it can't be changed, hence the error you're getting.

I'll implement a fix as soon as possible. In the meantime, the only workaround I can think of is to disable the default handler of Loguru through `os.environ["LOGURU_AUTOINI"]` and re-add it with a context explicitly:

```python
import os
import sys

os.environ["LOGURU_AUTOINIT"] = "0"

from sanic import Sanic, text
from loguru import logger

logger.add(sys.stderr, context="spawn")  # Assuming Sanic tries to set the start method to "spawn".

app = Sanic(__name__)


@app.get("/")
async def run_shell(request):
    return text("Hello world!")


if __name__ == "__main__":
    app.run(host="0.0.0.0", port=6666, debug=True, access_log=True, workers=2)
```

PR #983, opened 2023-09-10
Don't set system default multiprocessing.context

ZivRonen:
Create a new context using either the default method or the already set start method in multiprocessing.get_context, to avoid setting the default context while initializing the default handler.

Fix: https://github.com/Delgan/loguru/issues/974

### DIFF
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -92,7 +92,7 @@ import sys
 import warnings
 from collections import namedtuple
 from inspect import isclass, iscoroutinefunction, isgeneratorfunction
-from multiprocessing import current_process, get_context
+from multiprocessing import current_process, get_all_start_methods, get_context, get_start_method
 from multiprocessing.context import BaseContext
 from os.path import basename, splitext
 from threading import current_thread
@@ -967,8 +967,13 @@ class Logger:
         if not isinstance(encoding, str):
             encoding = "ascii"
 
-        if context is None or isinstance(context, str):
+        if isinstance(context, str):
             context = get_context(context)
+        elif context is None:
+            start_method = get_start_method(allow_none=True)
+            if start_method is None:
+                start_method = "fork" if "fork" in get_all_start_methods() else "spawn"
+            context = get_context(start_method)
         elif not isinstance(context, BaseContext):
             raise TypeError(
                 "Invalid context, it should be a string or a multiprocessing context, "


ZivRonen:
Need to fix macos and coverage, will created a new pull request afterward

Delgan:
@ZivRonen  Thanks for working on this!

However, I have a slightly different patch planned. I'll just pass `context` as `None` to the `Handler` and use `multiprocessing` directly there.

The problem with your proposed approach is that there is no reliable way to know what is the actual default start method of a system. For example, your implementation would use `"fork"` on MacOS while the default is actually `"spawn"`.

I also need to add unit tests for that and it will be a bit tricky (I'll need to start a different script in a new process so that global state isn't modified).

It's a critical bug, I'm very aware of that. I plan to publish a new release tonight or tomorrow. But I'm also taking the opportunity to check compatibility with Python 3.12 before releasing `v0.7.2` of Loguru.

ZivRonen:
Ok, thanks for the update.

However, when using multiprocessing directly, it still can set the default start method.
```
import multiprocessing
multiprocessing.get_start_method(allow_none=True)  # return None
multiprocessing.Queue()
multiprocessing.get_start_method(allow_none=True)  # return 'fork' on Linux 
```

Delgan:
Yes, good point. However, it won't a problem because `multiprocessing` won't be used if `enqueue=False` (the default).

The regression introduced  by Loguru `v0.7.1` is that adding a handler has also the side-effect of setting the start method, even if `multiprocessing` primitives are not required by the handler. Because Loguru automatically adds a `sys.stderr` handler, it breaks the code of anyone trying to call `set_start_method("spawn")` on Linux.

This is why I plan to pass `context=None` to the `Handler`, and the `Handler` will call `multiprocessing` only if `enqueue=True`. That means `logger.add(..., enqueue=True)` must be called after any `multiprocessing.set_start_method()`, but it's always been that way (except if Python changed the way using `multiprocessing` automatically set the default context).

ZivRonen:
Sound good. Thank you both for the great work and the in depth explanation.

Delgan:
Upon further reflection, I've come to understand your point more clearly. Your proposed approach indeed holds an advantage in that it refrains from setting the global start method. By consistently employing a `Context` created from `get_context()` with an explicit start method, we ensure that we do not unintentionally modify the application's global state, which aligns with best practices.

Nonetheless, I do have two primary concerns:

1. The `multiprocessing` module, to the best of my knowledge, lacks a function to retrieve the default start method. While we know that on Linux it's "fork," on MacOS it's "spawn," and on Windows it's "spawn," there could potentially be other platforms with distinct defaults. This could pose issues if our assumed start method doesn't align with the default context of a user employing `multiprocessing.Process()`.

2. If a user initially calls `logger.add(..., enqueue=True)` and we initialize the `Handler` internals using the inferred `"fork"` context on Linux, and later the user invokes `set_start_method("spawn")` and begins using the `logger` in a `multiprocessing.Process()` with the "spawn" method, it won't trigger an error. However, this scenario could lead to unexpected behavior due to the incompatibility between the two different context types. If `logger.add(..., enqueue=True)` were to set the global start method, at least calling `set_start_method("spawn")` afterwards would raise a `RuntimeError`, alerting the user to the issue.

In any case, I appreciate your kind words and your valuable contribution. :+1:

PR #996, opened 2023-09-24
Fix "variable possibly unbound" errors

FastestMolasses:
Im importing this library into my project by the source code and was getting linting errors about these 2 possibly unbound variables. The simple fix is to initialize these variables with their starting values.


Delgan:
Hi @FastestMolasses. May I ask with linter are you using please? This project uses `ruff` and maybe there are additional rules I should enable.

FastestMolasses:
Looks like it's not the linter, but it's coming from Pylance, the language server I am using on VSCode.

Screenshot 2023-10-01 at 2 12 07 PM

Delgan:
Ok, thanks for the answer and improvement!

PR #1001, opened 2023-10-02
Improve startuptime: don't import IPython/ipykernal if not in jupyter

zakstucke:
Importing `Ipython/ipykernal` when running from terminal slows startup a fair bit on my machine. 

If in jupyter `IPython` will already be imported, so can skip the imports directly when not needed. 

```bash
time python -c "import loguru"
```

Before:
```
0.31s user 0.06s system 99% cpu 0.374 total
0.22s user 0.02s system 99% cpu 0.242 total
0.20s user 0.03s system 99% cpu 0.237 total
0.21s user 0.03s system 99% cpu 0.242 total
0.22s user 0.03s system 99% cpu 0.245 total
```

After:
```
0.04s user 0.01s system 97% cpu 0.053 total
0.05s user 0.00s system 98% cpu 0.051 total
0.04s user 0.01s system 98% cpu 0.053 total
0.05s user 0.00s system 98% cpu 0.054 total
0.04s user 0.01s system 98% cpu 0.054 total
```

### DIFF
--- a/loguru/_colorama.py
+++ b/loguru/_colorama.py
@@ -6,7 +6,7 @@ def should_colorize(stream):
     if stream is None:
         return False
 
-    if stream is sys.stdout or stream is sys.stderr:
+    if "IPython" in sys.modules and (stream is sys.stdout or stream is sys.stderr):
         try:
             import ipykernel
             import IPython
--- a/loguru/_colorama.py
+++ b/loguru/_colorama.py
@@ -1,3 +1,4 @@
+import builtins
 import os
 import sys
 
@@ -6,7 +7,7 @@ def should_colorize(stream):
     if stream is None:
         return False
 
-    if "IPython" in sys.modules and (stream is sys.stdout or stream is sys.stderr):
+    if getattr(builtins, "__IPYTHON__", False) and (stream is sys.stdout or stream is sys.stderr):
         try:
             import ipykernel
             import IPython
--- a/tests/test_colorama.py
+++ b/tests/test_colorama.py
@@ -179,6 +179,16 @@ def test_jupyter_fixed(monkeypatch, patched, out_class, expected):
     assert should_colorize(stream) is expected
 
 
+def test_jupyter_missing_lib(monkeypatch):
+    # Missing ipykernal so jupyter block will err, should handle gracefully
+    stream = StreamIsattyFalse()
+
+    monkeypatch.setitem(sys.modules, "IPython", MagicMock())
+    monkeypatch.setattr(sys, "stdout", stream)
+
+    assert should_colorize(stream) is False
+
+
 @pytest.mark.parametrize("patched", ["__stdout__", "__stderr__"])
 @pytest.mark.skipif(os.name == "nt", reason="Colorama is required on Windows")
 def test_dont_wrap_on_linux(monkeypatch, patched, patch_colorama):
--- a/tests/test_colorama.py
+++ b/tests/test_colorama.py
@@ -1,3 +1,4 @@
+import builtins
 import os
 import sys
 from unittest.mock import MagicMock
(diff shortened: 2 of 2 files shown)


Delgan (commented):
You'll also need to change the unit tests for the coverage check to pass.

Two modifications are probably required:
* Update [`test_jupyter_fixed()`](https://github.com/Delgan/loguru/blob/14da892dceeaf3d5cc55b43de5e65c61bb795c4b/tests/test_colorama.py#L161-L179) and add `monkeypatch.setitem(builtins, "__IPYTHON__", True)`.
* Right below `test_jupyter_fixed()`, add a new `test_jupyter_missing_lib()` function which mocks `__IPYTHON__` and `sys.stdout` but without monkeypatching `sys.modules`. The idea is to force an `ImportError` to be raised in `should_colorize()` and ensure it's properly hanlded.

Delgan:
Great suggestion, thanks!

zakstucke:
@Delgan I think i've fixed the coverage with something similar to your `test_jupyter_missing_lib()` suggestion.

But the change from `sys.modules` is a bit tricker, I just tried it out and the mocking in tests doesn't seem to work.

`monkeypatch.setitem(builtins, "__IPYTHON__", True)` leads to:
`AttributeError: module 'builtins' has no attribute 'get'`

Similar with `setitem`: `AttributeError: <module 'builtins' (built-in)> has no attribute '__IPYTHON__'`

Any ideas? I personally don't think there's an issue leaving with the `sys.modules` check, the load cost is actually importing the packages, so if its already loaded in, the actual check itself is quick.

Delgan:
@zakstucke Thanks for the update regarding code coverage, and sorry for the delay.


> But the change from `sys.modules` is a bit tricker, I just tried it out and the mocking in tests doesn't seem to work.
> 
> `monkeypatch.setitem(builtins, "__IPYTHON__", True)` leads to: `AttributeError: module 'builtins' has no attribute 'get'`
> 
> Similar with `setitem`: `AttributeError: <module 'builtins' (built-in)> has no attribute '__IPYTHON__'`
> 
> Any ideas? I personally don't think there's an issue leaving with the `sys.modules` check, the load cost is actually importing the packages, so if its already loaded in, the actual check itself is quick.

Apologies, I made a msitake by suggesting `monkeypatch.setitem()`, it should have been [`monkeypatch.setattr()`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.MonkeyPatch.setattr). Because the `builtins.__IPYTHON__` variable does not exist by default, it indeed raises an `AttributeError`. However, this behavior can be disabled by specifying `raising=False`.

I agree that checking `sys.modules` is an equally appropriate solution. However, when possible I prefer not to rely on it. It's not impossible for the user to call `sys.modules.clear()` for some reason, therefore the solution relyong on `builtins` seems more robust to me (as it's not possible to inadvertently remove `__IPYTHON__`).
However it's certainly not a problem in practice, I'm mostly nitpicking.

I took the liberty of pushing a commit on your personal branch to update the PR. I wanted to check that it was working properly and not bother you any further.
Are you OK with the state of the PR?
If so, I'll merge it.

i-newton:
Hi guys!
Could you please tell me when this feature is going to be released? It saves us ~20Mb of RAM from every app that uses loguru.
Seems like important feature to release asap.

Delgan:
@i-newton Sorry, I have no ETA for now, but you can use install and the latest version:

```bash
pip install git+https://github.com/loguru/master.git@master
```

i-newton:
@Delgan Thanks for the info! Unfortunately, loguru version must be fixed in my current project due to project policy. That's why I am looking forward for version 0.7.3. Otherwise, i will have to fork the project (which is nor right option imo)

Delgan:
@i-newton You can pin a specific hash commit as well:

```bash
pip install git+https://github.com/loguru/master.git@bd66b55094c4f431e88dd0091f2b8b84fceee624
```

I hope to be able to release 0.7.3 in the near future, but I can't guarantee anything.

i-newton:
@Delgan sorry, for bothering you, do you still plan 0.7.3 release in near future?  unfortunately, i can't use commit tag due to customer policy

PR #1034, opened 2023-11-30
enable ruff/RET and fixing `RET505` Unnecessary elif after return statement

Borda:
As it seems Ruff is well used, I may suggest also enabling RET simplification and also fixing all raised issues of _`RET505` Unnecessary elif after return statement_

### DIFF
--- a/loguru/_colorizer.py
+++ b/loguru/_colorizer.py
@@ -238,10 +238,9 @@ class AnsiParser:
                     self._tokens.append((TokenType.CLOSING, "\033[0m"))
                     self._tokens.extend(self._color_tokens)
                     continue
-                elif tag in self._tags:
+                if tag in self._tags:
                     raise ValueError('Closing tag "%s" violates nesting rules' % markup)
-                else:
-                    raise ValueError('Closing tag "%s" has no corresponding opening tag' % markup)
+                raise ValueError('Closing tag "%s" has no corresponding opening tag' % markup)
 
             if tag in {"lvl", "level"}:
                 token = (TokenType.LEVEL, None)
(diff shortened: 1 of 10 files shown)


Borda:
I think that the codecov is a false alarm as we just reduced the number of lines os if any were not covered before, now they are just sending the wrong signal...

Delgan:
Hi @Borda, thanks for suggesting additional rules and implementing the fixes!

Regarding the Codecov failure: Loguru aims for 100% code coverage, and I think we could and should add a new unit test to maintain this standard.

[One new line](https://github.com/Delgan/loguru/pull/1034/files#diff-acab8b47700c70707c332aba6b5e85925db903d112d5a148251ce791545d0d6cR27) was added but is not covered.

Would you mind adding an unit test in [this file](https://github.com/Delgan/loguru/blob/0f0cd6172a953e295fe941b65bb49d1649e46b4d/tests/test_defaults.py) such as `test_invalid_type()` for example, and ensuring the `env()` function returns `None` in this case?

On a second thought, I also think we shouldn't return `None` but raise an error instead:

```python
raise ValueError("The requested type '%r' is not supported" % type_)
```

Borda:
@Delgan added Error + test

BTW, I think would be good to add a pattern for the message to be sure what exception you are receiving in a particular test...

Delgan:
Thank you @Borda!

> BTW, I think it would be good to add a pattern for the message to be sure what exception you are receiving in a particular test... 🦩 could be enforced by ruff PT

Yes, I agree. Many tests were implemented at a time I wasn't aware of the `match` argument of `pytest.raises()`. Ideally, it should be fixed.

PR #1066, opened 2024-01-10
FIX: ExceptionFormatter infinite recursion traceback

### LINKED ISSUE #1044: Loguru infinite recursion traceback loop
- [ciscoconfparse2](https://github.com/mpenning/ciscoconfparse2) uses loguru in a [pretty simple](https://github.com/mpenning/ciscoconfparse2/issues/1) manner, decorating methods with `@logger.catch(reraise=True)`

My problem is if I hit a bug in certain methods, I get an infinite traceback from loguru.  I love loguru when it works, but digging into the infinite recursion is a serious drag on my development because even with a multi-thousand-line scrollback buffer, I still can't hit control-c fast enough to interrupt the infinite loop of loguru messages and find the *real* cause of the problem (i.e. I have two problems, loguru inifinte recursion and the ciscoconfparse2 bug that caused it).

Is there a way to get loguru to stop printing tracebacks in an infinite loop?

You can reproduce the infinite recursion loop as follows:

- Install `loguru==0.7.2`
- git clone ciscoconfparse2 to my local system
- `cd ciscoconfparse2`
- checkout hash [`1a7b88d72c5a86a673ba8d4ed18316036c65fded`](https://github.com/mpenning/ciscoconfparse2/commit/1a7b88d72c5a86a673ba8d4ed18316036c65fded)
- run this script (python versions don't seem to matter... I have seen this problem over and over for months with a variety of python versions):

```python
from loguru import logger

from ciscoconfparse2 import CiscoConParse

parse = CiscoConfParse("tests/fixtures/configs/sample_01.junos", syntax="junos", factory=False)
print(parse)
for obj in parse.objs:
    print(obj)
```

ismagilli:
This is my attempt to fix #1044. I'm not sure how good the choice of variable names is, but I couldn't think of a better one. If there are any comments, I am ready to correct them.

### DIFF
--- a/loguru/_better_exceptions.py
+++ b/loguru/_better_exceptions.py
@@ -118,6 +118,10 @@ class SyntaxHighlighter:
             return
 
 
+class ExceptionFormatterRecursionError(Exception):
+    pass
+
+
 class ExceptionFormatter:
     _default_theme = {
         "introduction": "\x1b[33m\x1b[1m{}\x1b[0m",
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -97,7 +97,7 @@ from os.path import basename, splitext
 from threading import current_thread
 
 from . import _asyncio_loop, _colorama, _defaults, _filters
-from ._better_exceptions import ExceptionFormatter
+from ._better_exceptions import ExceptionFormatter, ExceptionFormatterRecursionError
 from ._colorizer import Colorizer
 from ._contextvars import ContextVar
 from ._datetime import aware_now
@@ -1219,6 +1219,7 @@ class Logger:
         class Catcher:
             def __init__(self, from_decorator):
                 self._from_decorator = from_decorator
+                self._already_logging_exception = False
 
             def __enter__(self):
                 return None
@@ -1239,8 +1240,13 @@ class Logger:
                 if from_decorator:
                     depth += 1
 
-                catch_options = [(type_, value, traceback_), depth, True] + options
-                logger._log(level, from_decorator, catch_options, message, (), {})
+                if not self._already_logging_exception:
+                    self._already_logging_exception = True
+                    catch_options = [(type_, value, traceback_), depth, True] + options
+                    logger._log(level, from_decorator, catch_options, message, (), {})
+                    self._already_logging_exception = False
+                else:
+                    raise ExceptionFormatterRecursionError()
 
                 if onerror is not None:
                     onerror(value)


ismagilli:
The code coverage test does not pass among the checks, but I do not quite understand how to coverage this `raise` statement.

ismagilli:
At the moment, there are several checks errors. I've fixed them, but I'd like to hear the comments on the implementation first and then prepare the commit.

Delgan:
Hi @ismagilli, thanks a lot for your efforts to resolve this bug!

The solution you proposed is a clever way to circumvent the infinite recursion. However, I was able to craft a code example that exhibits an unexpected behavior:

```python
from loguru import logger
import sys

if __name__ == "__main__":

    def sink(_):
        raise ValueError

    logger.remove()
    logger.add(sink, catch=False)

    catcher = logger.catch(reraise=False)

    class Foo:
        def __repr__(self):
            with catcher:
                raise ValueError

    print("Before: ", catcher._already_logging_exception)  # False (expected).

    try:
        foo = Foo()
        repr(foo)
    except Exception:
        pass

    print("After: ", catcher._already_logging_exception)  # True (unexpected).

    logger.remove()
    logger.add(sys.stderr)

    # The "ValueError" won't be captured by Loguru, instead of that
    # a "ExceptionFormatterRecursionError" will immediately be raised
    # and the program will exit.
    with catcher:
        raise ValueError
```

This is because if `logger._log()` raises an `Exception`, then `self._already_logging_exception = False` is never called.

You call easily fix it by adding a `finally` clause:

```diff
diff --git a/loguru/_logger.py b/loguru/_logger.py
index 84ac4cb..c26b6c5 100644
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -1242,9 +1242,11 @@ class Logger:
 
                 if not self._already_logging_exception:
                     self._already_logging_exception = True
-                    catch_options = [(type_, value, traceback_), depth, True] + options
-                    logger._log(level, from_decorator, catch_options, message, (), {})
-                    self._already_logging_exception = False
+                    try:
+                        catch_options = [(type_, value, traceback_), depth, True] + options
+                        logger._log(level, from_decorator, catch_options, message, (), {})
+                    finally:
+                        self._already_logging_exception = False
                 else:
                     raise ExceptionFormatterRecursionError()
 ```

Unfortunately, there is another problem that can't be fixed as easily. The following code still causes infinite recursion:

```python
from loguru import logger

if __name__ == "__main__":

    class Foo:
        def __repr__(self):
            with logger.catch(reraise=True):
                raise ValueError

    foo = Foo()
    repr(foo)
```

Since a new `Catcher` instance is created each time `__repr__` is called, there is no way to detect the recursion with the suggested implementation. :/

There are also thread-safety concerns. I don't have a suitable solution in mind yet. It's a tricky problem to solve.

ismagilli:
I completely agree with you. It only comes to mind to try to identify recursion by looking for repetitions in the call stack. But this will lead to false positives and false negatives.

ismagilli:
Since the above code does not fix the existing problem and I have no idea how to fix the code, there is no need to leave this pull request open.

Delgan:
Thanks for giving it a shot anyway. I may use your solution as inspiration for finalizing the fix.

Issue #1064, opened 2024-01-08
[Question] Is there a way to show logs on screen on windows for multiprocessing?


monchin:
Hi, I have read the document and have known that on windows, if I want to use multirprocessing, I need 
```python
logger.remove()
logger.add("file.log", enqueue=True)
```
But in this way, I need to continuously open `file.log` manually to see what has happened, which is a little unconvinient. So I wonder if there is any way to show logs on screen on windows even I use multiprocessing?

bojobo:
Have you tried logging to `stderr` instead of a file?

```py
import sys
...
logger.remove()
logger.add(sys.stderr, enqueue=True)

```

monchin:
> Have you tried logging to `stderr` instead of a file?
> 
> ```python
> import sys
> ...
> logger.remove()
> logger.add(sys.stderr, enqueue=True)
> ```

I haven't, because I saw [the document](https://loguru.readthedocs.io/en/stable/resources/recipes.html#compatibility-with-multiprocessing-using-enqueue-argument), and it said that `Default "sys.stderr" sink is not picklable`, I think this means when using multiprocessing on windows, we can't use loguru with `sys.stderr`

Delgan:
@monchin Both `sys.stderr` and `"file.log"` handlers aren't picklable by default. To use them properly with `multiprocessing`, they need to be configured with `enqueue=True`. When `enqueue=True`, they became picklable. 

The documentation refers to the default `sys.stderr` sink pre-configured when importing Loguru. This one was added with `enqueue=False` and therefore can't be pickled. However, as @bojobo suggested, using `logger.add(sys.stderr, enqueue=True)` is perfectly fine and should resolve your problem.

I think the documentation is inaccurate, instead of `Default "sys.stderr" sink is not picklable`  it should be "Default "sys.stderr" hander is not picklable".

monchin:
> @monchin Both `sys.stderr` and `"file.log"` handlers aren't picklable by default. To use them properly with `multiprocessing`, they need to be configured with `enqueue=True`. When `enqueue=True`, they became picklable.
> 
> The documentation refers to the default `sys.stderr` sink pre-configured when importing Loguru. This one was added with `enqueue=False` and therefore can't be pickled. However, as @bojobo suggested, using `logger.add(sys.stderr, enqueue=True)` is perfectly fine and should resolve your problem.
> 
[...]

Issue #912, opened 2023-07-03
Proper way to use `loguru.logger` with `multiprocessing`( on Mac M1)


changchiyou:
## My env

- Apple Mac M1 Air
- 8GB RAM
- Ventura 13.2.1（22D68）

## Question 
1. Is there more example of using `loguru.logger` with `multiprocessing` or other multi-process library and shared the configuration of `loguru.logger`? I have read [Compatibility with multiprocessing using enqueue argument](https://loguru.readthedocs.io/en/stable/resources/recipes.html#compatibility-with-multiprocessing-using-enqueue-argument) in [Code snippets and recipes for loguru](https://loguru.readthedocs.io/en/stable/resources/recipes.html#code-snippets-and-recipes-for-loguru) but many error occured on my project after I followed that.

2. Can I pass the `loguru.logger` instance through methods as argument and reload/reset/overwrite the `loguru.logger` in sub-process which is created by "spawning" instead of "forking"(I got many warning message when I tried to use "forking" on Mac M1)?

## The error mentioned in `1.`, `2.`

The setting of `loguru.logger` missed after my sub-process is created:

```
# settings including two `logger.add()` for logging on console and recording into logs
set_loguru(
    enqueue=True,
    console_level=logging.DEBUG
    if config_setting.args.DEBUG is True
    else logging.INFO,
)

# this line work perfectly with the setting above (from `set_loguru`)
logger.debug("start")

# but the logger inside missed the setting
workers = []
for section_name, mark in config_setting.get_matching_sections():
    worker = multiprocessing.Process(
        target=realtime,
        args=(mark, section_name, self.config_path, self.args_dict),
        daemon=True,
    )
    worker.start()
    workers.append(worker)
```

![image](https://github.com/Delgan/loguru/assets/46549482/d519f6b4-4fc2-4300-9817-b54a477b3371)

I have tried to set 

```
multiprocessing.set_start_method("fork")
``` 
on the beginning of my project to avoid `loguru.logger` being recreated due to the spawning/forking differences mentioned in https://github.com/Delgan/loguru/issues/908#issuecomment-1615896444:

> if the child process is created by "spawning" and not by "forking", then two different logger objects will be created, and they won't share the same handlers.

[...]

Issue #818, opened 2023-03-12
child process (spawn context): set parent logger globally for all modules


PakitoSec:
Thanks for this awesome library.

I would like to set the parent logger globally in a child process (spawn method) so I don't have to do that within all my modules like I saw in this example: [loguru multiprocess](https://loguru.readthedocs.io/en/stable/resources/recipes.html#compatibility-with-multiprocessing-using-enqueue-argument).

The only way I found is by using the code below but I think this is not recommended since `_core` is not in the public api:

```python
Module A:
from loguru import logger
...
    def entrypoint_child_process(self, logger_):
        logger._core = logger_._core

Module B:
from loguru import logger
...
    def child_job(self):
        logger.info("test")
```

What could be the solution in order to set the arg `logger_` globally in the child process so I don't have to do the trick describe in your documentation in each module ?

UPDATE:
Maybe this could be accepted ?
```python
    logger.__dict__ = logger_.__dict__.copy()
```

PakitoSec:
no answer

Delgan:
Sorry for not answering earlier. The fact is that I didn't have a solution, and this ticket made me realize that there was room for improvement in the way `multiprocessing` is handled.

The `__dict__` update seems to be a working solution, but it's probably not very elegant.

I thought about it a little bit and I guess there is no other solution than to provide a new method to reload the logger from another one.

```python
from multiprocessing import Process
from loguru import logger

def entrypoint_child_process(self, logger_):
    logger.reload(logger_)
    logger.info("Test")

if __name__ == "__main__":
    logger.remove()
    logger.add("file.log", enqueue=True)
    process = Process(target=entry_child_process, args=(logger, ))
    process.start()
    process.join()
```

I'm also disappointed that the `logger` object must be passed as an argument to `Process()` or `map()` on Windows or when the `"spawn"` context is used, but I haven't found a viable alternative that would avoid doing this. I am open to suggestions.

PakitoSec:
I think you can't avoid logger object to be passed as an argument to multiprocessing with spawn context since you are losing the configuration by design. It's the same with std logger.

`_core` should be a singleton and could be updated with the logger object passed from Process/Pool...
`logger.__dict__ = logger_.__dict__.copy()` won't work for logger created from `logger.bind` in the child process.

Your suggestion to create a reload logger method for this purpose seems good.

Delgan:
[...]

PR #1069, opened 2024-01-20
Add "reinstall()" method to make it easier in spawn multiprocessing 

monchin:
Make it easier when using `loguru` in spawn multiprocessing environment

### DIFF
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -86,6 +86,7 @@ import builtins
 import contextlib
 import functools
 import logging
+import os
 import re
 import sys
 import warnings
@@ -230,6 +231,7 @@ class Logger:
     def __init__(self, core, exception, depth, record, lazy, colors, raw, capture, patchers, extra):
         self._core = core
         self._options = (exception, depth, record, lazy, colors, raw, capture, patchers, extra)
+        self._own_pid = os.getpid()
 
     def __repr__(self):
         return "<loguru.logger handlers=%r>" % list(self._core.handlers.values())
@@ -1751,6 +1753,46 @@ class Logger:
                     self.disable(name)
 
         return [self.add(**params) for params in handlers]
+
+    def _replace_core(self, core: Core):
+        self._core = core
+
+    def reinstall(self):
+        """Reinstall the core of logger.
+
+        When using multiprocessing, you can pass logger as a parameter to the target of
+        ``multiprocessing.Process``, and run this method once, thus you don't need to pass logger to every
+        function you called in the same process with spawn multiprocessing.
+
+        Examples
+        --------
+        >>> def subworker(logger_):
+        ...     logger_.reinstall()
+        ...     logger.info("Child")
+        ...     deeper_subworker()
+
+        >>> def deeper_subworker():
+        ...     logger.info("Grandchild")
+
+        >>> def test_process_spawn():
+        ...     spawn_context = multiprocessing.get_context("spawn")
+        ...     logger.add("file.log", context=spawn_context, enqueue=True, catch=False)
+        ...
+        ...     process = spawn_context.Process(target=subworker, args=(logger,))
+        ...     process.start()
+        ...     process.join()
+
+        ...     assert process.exitcode == 0
+
+        ...     logger.info("Main")
+        ...     logger.remove()
+        """
+        if self._own_pid == os.getpid():  # same process
+            return
+        from loguru import logger
+
+        logger._replace_core(self._core)
+
 
     def _change_activation(self, name, status):
         if not (name is None or isinstance(name, str)):
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -86,7 +86,6 @@ import builtins
 import contextlib
 import functools
 import logging
-import os
 import re
 import sys
 import warnings
@@ -231,7 +230,6 @@ class Logger:
     def __init__(self, core, exception, depth, record, lazy, colors, raw, capture, patchers, extra):
         self._core = core
         self._options = (exception, depth, record, lazy, colors, raw, capture, patchers, extra)
-        self._own_pid = os.getpid()
 
     def __repr__(self):
         return "<loguru.logger handlers=%r>" % list(self._core.handlers.values())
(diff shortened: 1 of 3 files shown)


monchin:
It seems `black` does not check docstring's length.

Delgan:
Thanks for the PR @monchin!

However, I'm going to wait a little before merging it, because there are quite a few special cases I'd like to think about and a lot of unit tests that need to be implemented. I'll probably make some commits after yours to complete the development, if you're OK with that.

> It seems `black` does not check docstring's length.

Indeed, docstrings are not automatically formatted by `black`, we have to do it manually.

loguru/_logger.py
```diff
@@ -1752,6 +1754,46 @@ def configure(self, *, handlers=None, levels=None, extra=None, patcher=None, act
...
+        ...     process = spawn_context.Process(target=subworker, args=(logger,))
+        ...     process.start()
+        ...     process.join()
+
+        ...     assert process.exitcode == 0
+
+        ...     logger.info("Main")
+        ...     logger.remove()
+        """
+        if self._own_pid == os.getpid():  # same process
```
Delgan, line 1790: I don't think we should check for the PID. For example, if someone "clone" the `logger` with `copy.deepcopy()` ([see docs](https://loguru.readthedocs.io/en/stable/resources/recipes.html#creating-independent-loggers-with-separate-set-of-handlers)) and then call `copied_logger.reinstall()`, I believe it should replace the `_core` of the default `logger` even within the same process.

loguru/_logger.py
```diff
@@ -1752,6 +1754,46 @@ def configure(self, *, handlers=None, levels=None, extra=None, patcher=None, act
...
+            return
+        from loguru import logger
+
+        logger._replace_core(self._core)
```
Delgan, line 1794: I think we can remove the `_replace_core()` and do `logger._core = self._core` directly.

tests/test_multiprocessing.py
```diff
@@ -209,7 +219,7 @@ def test_process_spawn(spawn_context):
 
     logger.add(writer, context=spawn_context, format="{message}", enqueue=True, catch=False)
 
-    process = spawn_context.Process(target=subworker, args=(logger,))
+    process = spawn_context.Process(target=subworker_spawn, args=(logger,))
```
Delgan, line 222: I would prefer we do not touch the existing unit tests. Could you please create a new `tests/test_reinstall.py` file where we'll put all tests we need?

loguru/_logger.py
```diff
@@ -1752,6 +1754,46 @@ def configure(self, *, handlers=None, levels=None, extra=None, patcher=None, act
...
+        """Reinstall the core of logger.
+
+        When using multiprocessing, you can pass logger as a parameter to the target of
+        ``multiprocessing.Process``, and run this method once, thus you don't need to pass logger to every
+        function you called in the same process with spawn multiprocessing.
+
+        Examples
+        --------
+        >>> def subworker(logger_):
+        ...     logger_.reinstall()
```
Delgan, line 1770: I think it would be more convenient to name the argument `logger` instead of `logger_`. Then we can do:

```python
logger.reinstall()
logger.info("Child")
```

It should work transparently.

loguru/_logger.py
monchin: Yes, you're right. I'll delete it.

loguru/_logger.py
monchin: Actually I also like `logger._core = self._core` directly, but I saw your comment in [#818](https://github.com/Delgan/loguru/issues/818#issuecomment-1538688196) so I thought you prefer this way XD. I'll change it.

monchin:
I'm not familier with `ruff`, seems it failed because I created a new `tests/test_reinstall.py` file

Delgan:
Thanks for the update.



> I'm not familiar with `ruff`, seems it failed because I created a new `tests/test_reinstall.py` file

Similarly to `black`, the `ruff` tool notably replaces `isort` and is in charge of re-ordering the `import` statements.

You can try to install the pre-commit hooks to automatically fix your files before `git commit`:

```bash
$ pre-commit install --install-hooks
```

Or you can manually fix the file with:

```bash
$ ruff tests/test_reinstall.py --fix
```

indigoviolet:
@Delgan What is the status of this? this would be a really nice addition to have.

Delgan:
Thanks for your contribution and sorry for the delay.

The `logger.reinstall()` method is a great addition. I'll merge the PR as-is and see if further refinements are needed (thinking about some edge cases).

Refs: #818 #912 #1064

wynemo:
hello, guys, does it actually work?

run it on linux, it fails:

```
import multiprocessing
from loguru import logger

def subworker(_logger):
    _logger.reinstall()
    _logger.info("Child")


def test_process_spawn():
    spawn_context = multiprocessing.get_context("spawn")
    logger.add("file.log", context=spawn_context, enqueue=True, catch=False)

    process = spawn_context.Process(target=subworker, args=(logger,))
    process.start()
    process.join()
    assert process.exitcode == 0
    _logger.info("Main")
    _logger.remove()

if __name__ == "__main__":
    test_process_spawn()

ubuntu@oracle:/tmp$ ./venv/bin/python test.py
  File "/tmp/test.py", line 21, in <module>
    test_process_spawn()
  File "/tmp/test.py", line 14, in test_process_spawn
    process.start()
  File "/usr/lib/python3.10/multiprocessing/process.py", line 121, in start
    self._popen = self._Popen(self)
  File "/usr/lib/python3.10/multiprocessing/context.py", line 288, in _Popen
    return Popen(process_obj)
  File "/usr/lib/python3.10/multiprocessing/popen_spawn_posix.py", line 32, in __init__
    super().__init__(process_obj)
  File "/usr/lib/python3.10/multiprocessing/popen_fork.py", line 19, in __init__
    self._launch(process_obj)
  File "/usr/lib/python3.10/multiprocessing/popen_spawn_posix.py", line 47, in _launch
    reduction.dump(process_obj, fp)
  File "/usr/lib/python3.10/multiprocessing/reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
TypeError: cannot pickle '_io.TextIOWrapper' object
```

am i missing something?

monchin:
> Thanks for your contribution and sorry for the delay.

Thank you for merging and it's my pleasure.

I wanted to make loguru name rotation files the same way as std logging when rotating every midnight, and found that actually I can use `logger.add(f"file.log.{time:YYYY-MM-DD})`. Now I can replace std logging with loguru from next release！

Delgan:
@wynemo The new `reinstall()` method only helps with updating the global `logger` after it's inherited by a child process. However, this does not change the fact that the `logger` must be inheritable to begin with.

In your case, you're facing an error because the default sink, which is `sys.stderr`, can't be inherited (because it contains non-picklable attributes). This is a common problem. You need first to remove the default sink, then re-add it with `enqueue=True` (which ensures the sink can be inherited):

```python
spawn_context = multiprocessing.get_context("spawn")
logger.remove()
logger.add(sys.stderr, context=spawn_context, enqueue=True)
logger.add("file.log", context=spawn_context, enqueue=True, catch=False)

```

Issue #1072, opened 2024-01-27
Help designing an integration with `logot`


etianen:
I've recently released a 0.1.0 preview of a log testing library called [logot](https://github.com/etianen/logot). It provides a nice fixture for testing [complex log patterns](https://logot.readthedocs.io/latest/logged.html), which I've found very useful when testing multithreaded and async codebases.

I'd like to provide a `loguru` integration. 🤗Since I'm not hugely familiar with `loguru`, I was hoping you'd have some advice on what a good integration would look like.

There are a couple of sticking points that I'd like to nail before committing to a v1.0.0 API.

# Log capture

Currently, stdlib logging capture is enabled with a [context manager method](https://logot.readthedocs.io/latest/api.html#logot.Logot.capturing):

``` py
from logot import Logot

with Logot().capturing() as logot:
    pass
```

Since the `loguru` integration would be an optional `extra`, a freestanding context manager could be provided like this:

``` py
from logot.contrib.loguru import capturing

with capturing(Logot()) as logot:
   ...
```

It's a bit awkward having a freestanding context manager for 3rd party integrations, so an alternative would be a capturing backend approach:

``` py
from logot.contrib.loguru import LoguruCapture

with Logot().capturing(LoguruCapture):
   pass
```

Any arguments given to `capturing` would be propagated to the `LoguruCapture` constructor. This would then unify capture between stdlib logging and `loguru` logging. The default capture backend would be the stdlib capture, and an `@overload` would ensure good typing. 💪

The alternative is a whole `Logot` subclass just for `loguru`, which seems heavyweight and unpleasant. 🤮

## Log message matching

Rather than awkward regexes, `logot` lets you [match log messages](https://logot.readthedocs.io/latest/match.html) with `%` placeholders. The hope here is that it's much cleaner and more intuitive to match using the syntax used for log message placeholders. (Similar to how `printf` and `scanf` pair nicely).

``` py
from logot import logged

logot.wait_for(logged.info("Hello %s"))
```

But `loguru` uses lovely `.format()`-style placeholders!

Maybe it's fine to just use `%` placeholders in `loguru` tests? 🤷

But it might be nice to support `.format()`-style placeholders like this:

``` py
from logot.contrib.loguru import logged

logot.wait_for(logged.info("Hello {}"))
```

Support for matching structured data would also be possible:

``` py
logot.wait_for(
    logged.bind(foo="bar").info("Hello {}")
)
```

## Advice appreciated! 🙏

[...]

PR #1085, opened 2024-02-20
Linking to `logot` from docs

etianen:
I've added a new "Testing logging" section to the "Code snippets and recipes" doc. There didn't seem to be any particular order here, and placing it in an arbitrary place felt... _arbitrary_, while the top felt... _presumptuous_! I've linked to the existing migrations guides for existing projects with existing test cases.

I've also added a link _to_ "Testing logging" from the migrations guides. Hopefully that cross-links things nicely without blarting all over your docs.

Let me know if there's anything you'd like changed! 🙇 

See #1072

### DIFF
--- a/docs/resources/migration.rst
+++ b/docs/resources/migration.rst
@@ -278,6 +278,8 @@ The |captureWarnings| function which redirects alerts from the |warnings| module
     warnings.showwarning = showwarning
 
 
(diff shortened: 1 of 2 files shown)


Delgan:
That looks good, thank you @etianen!

PR #1106, opened 2024-03-22
Modified Internal log level cache for better interoperability

misotrnka:
In my work I've come accross situations, where a third-party package is using my Loguru logger to emit logs. I've noticed that many time these messages are formated (colored) differently from messages emited in my code. After a brief investigation it turned out the third-party package is calling the `log()` method like this:

```python
self.logger.log(logging.INFO, message)
```

The `logging.INFO` variable is an integer constant `20`. However, Loguru's `Logger._log()` method expects a string, such as `"INFO"`. Not finding the integer `20` in the internal `core.levels_lookup` cache results into a different formatting of the message:

image

(The first message is logged through the third-party package, while the second is logged natively in my code)

It is not possible to solve this using a custom log filter, because in Loguru the log level is resolved internally and the `LogRecord.levels` information is not used. 

I believe passing the log level using the logging package's enum is a good and wide-spread practice and should be supported by Loguru.

**I propose a simple solution:** adding integer keys in additon to string keys to the internal log level cache. This effectivelly resolves the issue, increasing Loguru's interoperability with other packages without any modifications the the internal processing flow and logic.

### DIFF
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -178,9 +178,11 @@ class Core:
         # Cache used internally to quickly access level attributes based on their name or severity.
         # It can also contain integers as keys, it serves to avoid calling "isinstance()" repeatedly
         # when "logger.log()" is used.
-        self.levels_lookup = {
-            name: (name, name, level.no, level.icon) for name, level in self.levels.items()
-        }
+        self.levels_lookup = {}
+        for name, level in self.levels.items():
+            # Update the dictionary for both keys - the log level name and the log level severity number
+            self.levels_lookup[name] = (name, name, level.no, level.icon)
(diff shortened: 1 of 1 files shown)


while0pass:
> Basically, it means the users of your library should use `"INFO"` instead of `logging.INFO`, since this is the expected usage of Loguru.

I would expect that in the loguru settings there should be an option to specify the mapping of numeric levels to expected level names for specific libraries. For loguru users, this could be a relatively convenient way to solve the current problem (I've encountered it too and the current state frustrates me, as it contradicts the idea of bringing more ease to logging). With this solution, there is no need for both abandoning loguru's advanced severity level approach and the unrealistic requirement for all Python library creators to use only level names when logging.

PR #1136, opened 2024-05-04
Publish a new package, loguru-hardened, with security focussed defaults

bneijt:
Introduce a separate script that will override a few files to harden loguru defaults and build a loguru-hardened package, then revert the changes using git.


hardened/_defaults.py
Delgan, line 1: Can we avoid duplicating this file? What about a `HARDENED = True` variable somewhere, then simply do something like this in the existing `_default.py` file:

```python
LOGURU_DIAGNOSE = env("LOGURU_DIAGNOSE", bool, False if HARDENED else True)
```

build_hardened.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # Replace loguru with loguru-hardened
+    setup_py = setup_py.replace('name="loguru"', 'name="loguru-hardened"')
+
+    # Write the updated setup.py file
+    with open("setup.py", "w") as f:
+        f.write(setup_py)
+
+
+def replace_with_hardened_files():
```
Delgan, line 19: Similar to what I suggested in my other comment, maybe we can simplify the process by simply having a `HARDENED` boolean value somewhere, and switch it from `False` to `True` for the `loguru-hardened` build?

Delgan:
I don't know why 3.13 build fails, maybe try rebasing (a few libraries were updated since then)?

trim21:
if it's serialized be default, should we also remove level time from default format?

build_hardened.py
bneijt: Think that's a great solution, I wanted to keep away from patching python, but if we are only patching a single variable assignment, I think we should be fine.

loguru/_defaults.py
```diff
@@ -39,9 +40,9 @@ def env(key, type_, default=None):
 LOGURU_FILTER = env("LOGURU_FILTER", str, None)
 LOGURU_LEVEL = env("LOGURU_LEVEL", str, "DEBUG")
 LOGURU_COLORIZE = env("LOGURU_COLORIZE", bool, None)
-LOGURU_SERIALIZE = env("LOGURU_SERIALIZE", bool, False)
+LOGURU_SERIALIZE = env("LOGURU_SERIALIZE", bool, True if HARDENED_BUILD else False)
```
bneijt, line 43: We could just say HARDENED_BUILD, but I like the explicit notation of using an if to evaluate the true/false. Feel free to comment on the contrary.

bneijt:
> if it's serialized be default, should we also remove level time from default format?

I don't follow, could you give a code example of what you mean or explain further?

trim21:
> > if it's serialized be default, should we also remove level time from default format?
> 
> I don't follow, could you give a code example of what you mean or explain further?

from
```python
LOGURU_FORMAT = env(
    "LOGURU_FORMAT",
    str,
    "<green>{time:YYYY-MM-DD HH:mm:ss.SSS}</green> | "
    "<level>{level: <8}</level> | "
    "<cyan>{name}</cyan>:<cyan>{function}</cyan>:<cyan>{line}</cyan> - <level>{message}</level>",
)
```
to 
```python
LOGURU_FORMAT = env("LOGURU_FORMAT", str, "{message}")
```

README.rst
```diff
@@ -40,6 +40,7 @@ Installation
 
     pip install loguru
 
+Or if you need more secure defaults, install ``loguru-hardened``.
```
bneijt, line 43: I also mentioned `loguru-hardened` in the README, but I could not find a release action configured, so I'm not sure how to make sure hardened is also released when a new release is created.

loguru/_defaults.py
Delgan: I totally agree with you here. Explicitly conversion from `HARDENED_BUILD` to a boolean value is better. :+1:

README.rst
Delgan: Indeed, since there is no Github Action, I need to do it manually for now.

loguru/_defaults.py
```diff
@@ -27,21 +27,27 @@ def env(key, type_, default=None):
...
 
 LOGURU_FORMAT = env(
     "LOGURU_FORMAT",
     str,
-    "<green>{time:YYYY-MM-DD HH:mm:ss.SSS}</green> | "
-    "<level>{level: <8}</level> | "
-    "<cyan>{name}</cyan>:<cyan>{function}</cyan>:<cyan>{line}</cyan> - <level>{message}</level>",
+    (
+        "{message}"
+        if HARDENED_BUILD
```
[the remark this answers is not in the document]
bneijt: I feel like having ‘{message}’ should be added to create a sensible combination of defaults when hardened is enabled.

For structured logging, the message only format makes the most sense. The user is still able to change it and move away from the defaults is required.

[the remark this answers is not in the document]
Delgan: Fine. Let's use `"{message}"` by default for hardened build until I improve the serialized format for all.

Delgan:
I finally fixed CI for Python 3.5 and 3.13, can you try rebasing the PR please?

README.rst
bneijt: I could open up another pr with release workflows if you would like that. I don't know what currently goes into the manual release actions, but it might be a nice setup to simplify maintenance.

Delgan: @bneijt Thanks for the proposal. If you're familiar with release automation via Github Actions, your help is most welcome. I'd like to replace manual release management.

Also sorry for not having yet merged this PR, but it seems it will be incompatible with https://github.com/Delgan/loguru/pull/1140. I need to check how to create `loguru-hardened` without `setup.py`.

trim21: just use a Python script change project.name in new pyproject.toml file

trim21: i can help write a ci file

Delgan: @trim21 Yes, I suppose this should be straightforward. Unfortunately, this PR will need to be updated, though.

trim21:
I have a question about this, what if a package I'm using require `loguru`, and my application require this package, and use `loguru-hardened`? it will conflict

README.rst
trim21: consider merge 1140 first maybe....

bneijt:
> I have a question about this, what if a package I'm using require `loguru`, and my application require this package, and use `loguru-hardened`? it will conflict

Yes, there will be a conflict which needs to be solved by uninstalling or ignoring the loguru package in environments that require you to use a hardened version instead.

The proposed workaround of using an extra won’t work because the reason for having this package is to facilitate environments where static code analysis like sonarqube flag loguru for being difficult to use in a secure way.

Which means that we can’t have loguru installed.

Next week I will find some time to create a hood example of how to use loguru hardened in those situations and explain how to deal with oackages that require loguru.

bneijt:
> That's false positive problem of static code analysi themselve, right? It doesn't make sence if you have env var `LOGURU_SERIALIZE=True`, you won't (can't) have multiple line log message in json

Yes and no.. it is not a false positive that “the default don’t protect against line injection properly”, but you are right that that is a bad mark in the first place because that same reasoning should flag Python itself (logging has the same issue) ;)

trim21:
> Yes and no.. it is not a false positive that “the default don’t protect against line injection properly”, but you are right that that is a bad mark in the first place because that same reasoning should flag Python itself (logging has the same issue) ;)

So with extra, the default has been changed, right?

bneijt:
We switched to an open telemetry implementation, so I won't be able to move forward on migrating projects to loguru at this point and I also won't be able to do the Sonar Qube validation after potential release.

Apart from that, even if we create a hardened version, any package picking up the normal loguru will still result in the total environment being flagged.

I still think the project should consider adding continuation characters as a solution to allowing newlines while still making them visible if injected by a third party, but as I won't be able to test this I can't see me proceeding on this.

Closing the issue and PR.

PR #1140, opened 2024-05-13
build: use pep517/pep621

trim21:
replace legacy setup.py with pyproject.toml pep517/pep621 config


Delgan:
> would you mind enable this option in repo settings? so I can update branch in web-ui instead of local git client.

Done, thanks for the suggestion. :+1:

trim21:
I guess, ci failed because setuptools support python3.5/3.6 are very out-dated and doesn't support new feature we are using.

trim21:
at fitst I want to continue using setuptools but python3.5/3.6 are too old and setuptools support thesee 2 version missing some feature we need. So I have to use flit to make it work on these 2 versions

Delgan:
Hum... Reading documentation about Flir it seems it does not support build step which might be a problem in the future if part of Loguru is rewritten in C. :confused:

trim21:
> Hum... Reading documentation about Flir it seems it does not support build step which might be a problem in the future if part of Loguru is rewritten in C. 😕

as I mentioned, drop py3.5/3.6 and use setuptools

Delgan:
Yes, let's do it that way. I was planning to drop support for older versions of Python when the C version would be implemented anyway.
Assuming that ever happens...

Thanks for your work.

trim21:
no problem. I use this package a lot (almost every application project of mine), don't want to see it dead.

Delgan:
Glad to know that Loguru is important to you! I agree that Loguru's development lacks activity compared to what I was aiming for, unfortunately. I'm trying to get back to it progressively. Thanks for your help, it's much appreciated.

PR #1142, opened 2024-05-16
style: check file eol

trim21:


.pre-commit-config.yaml
```diff
@@ -5,6 +5,13 @@ repos:
   - id: end-of-file-fixer
   - id: trailing-whitespace
   - id: check-added-large-files
+  - id: mixed-line-ending
+    args: [--fix=lf]
+    exclude_types: [batch]
+  - id: mixed-line-ending
+    args: [--fix=crlf]
+    types: [batch]
+
```
Delgan, line 14: I don't think this repository will ever contain batch files, can we just force LF for every files please?

```suggestion
  - id: mixed-line-ending
    args: [--fix=lf]**
```

trim21: it does https://github.com/Delgan/loguru/blob/master/docs/make.bat

Delgan: Haha, my bad.

Delgan:
Thanks!

.pre-commit-config.yaml
Delgan: Yes, I agree. I should have run `sphinx-build` with the `--no-batchfile` argument.

PR #1146, opened 2024-06-01
ci: add pypi release workflow

trim21:
how to use this:

create a pypi api token at https://pypi.org/manage/account/ and set project secret at https://github.com/Delgan/loguru/settings/secrets/actions , with name `PYPI_API_TOKEN`. 

After it's done, change version and add a git tag, then push it to github.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -18,6 +18,8 @@ jobs:
 
     - run: pipx run build
 
+    - run: twine check --strict dist/*
+
     - run: pipx run twine upload dist/* --disable-progress-bar
       env:
         TWINE_USERNAME: __token__
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,24 @@
+name: release
+
+on:
+  push:
+    tags:
+    - '*'
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - uses: actions/setup-python@v5
+      with:
+        python-version: 3.x
+
+    - run: pipx run build
+
+    - run: pipx run twine upload dist/* --disable-progress-bar
+      env:
+        TWINE_USERNAME: __token__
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - uses: actions/setup-python@v5
+      with:
+        python-version: 3.x
+
+    - run: pipx run build
```
Delgan, line 19: Can we use `python -m build` directly instead of `pipx`? Also, can we please add a `twine check --strict dist/*` step?

Delgan (commented):
That's great, thanks for the quick PR.

.github/workflows/release.yaml
trim21: `build` is not a stdlib and need to be installed with `pip`, which is what `pipx run build` mean, pip install it and run it

Delgan: Well, I suppose we just need to install it first. I'll add it to the `tox.ini` commands. Thanks.

Delgan:
I'll merge it as is, but for now it can't be used yet. I need to make the Readme compatible (it's not validated by Twine, I used to manually change a few things).

.github/workflows/release.yaml
trim21: > Well, I suppose we just need to install it first. I'll add it to the `tox.ini` commands. Thanks.

you don't need to add it...

Delgan: I certainly don't need to, but I prefer to regroup the commands and dependencies in one-place.

webknjaz:
FYI https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

trim21:
> FYI https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

what's your point?

Issue #1174, opened 2024-07-15
Allow for combining file rotation triggers by accepting a list of rotation conditions


CollinHeist:
## Background
I found myself (and [others](https://github.com/Delgan/loguru/issues/241)) wanting to rotate log files based on a combination of two already-implemented conditions (file size and interval in my case). I see there is existing functionality to allow arbitrary logic in the rotation by passing in a generic `callable`, however that would have required me to re-implement much of loguru's existing rotator functionality (or import code from within the loguru source).

A cleaner approach would be to permit the specification of lists (or generic iterables) in the `rotation` argument, and then have loguru parse each element into a combined rotation rotation logic where a file will be rotated if _any_ of the rotators evaluate to true.

## Example

```py
from loguru import logger

logger.configure(
  handlers=[
    dict(
      sink='file.log',
      rotation=['12h', '10 MB'],
    )
  ]
)
```
Would rotate log files every 12 hours _or_ every 10 Megabytes. This is much easier for the end-user than creating a custom callable of the same logic.

I've submitted this [PR](https://github.com/Delgan/loguru/pull/1175)

Delgan:
Fixed by #1175.

PR #1175, opened 2024-07-15
Allow list of rotation conditions in file rotation logic

CollinHeist:
This implements the behavior described in #1174 , and allows for specification of a list (or tuple/set) of conditions for file rotation, like so:

```py
from loguru import logger

logger.configure(
  handlers=[
    dict(
      sink='file.log',
      rotation=['12h', '10 MB'],
    )
  ]
)
```


Delgan:
Thanks for your PR @CollinHeist. Apologies for the delay.

Admittedly, I was initially rather reluctant to add such a feature, for the sake of simplicity. Though, I agree that this is not actually very cumbersome to add, and indeed, and it can visibly be useful to several users.

Just to let you know that I haven't forgotten the PR and plan to merge it after releasing a new debug release. Thank you again for your work.

Issue #1183, opened 2024-07-31
RuntimeError when reinitializing logger


jeremyk:
We hit this error intermittently as we have to reinitialize our logger at least once during our application startup.
```
File "/opt/pysetup/.venv/lib/python3.11/site-packages/loguru/_logger.py", line 2031, in _log
    for handler in core.handlers.values():
RuntimeError: dictionary changed size during iteration
````

I think this can be fixed by adding a lock before this line:
https://github.com/Delgan/loguru/blob/master/loguru/_logger.py#L2036

**Comment by jeremyk** [comment]
Fixes https://github.com/Delgan/loguru/issues/1183

**Comment by horatiu-negutoiu** [comment]
+1

**Comment by ericscales** [comment]
Another +1

**Comment by jeremyk** [comment]
Just to sanity check that this is needed:
```
>>> x = {"a": "1", "b": "2"}
>>> y = x.values()
>>> z = list(x.values())
>>> x["c"] = "3"
>>> y
dict_values(['1', '2', '3'])
>>> z
['1', '2']
```

Delgan:
Thanks for the bug report. I'm sorry you've had trouble because of this.

The logger was designed so that adding or removing handlers could be done in parallel with logging, but there's a flaw in the current implementation. I'll make sure it's fixed.

Delgan:
Fixed on `master` branch.

rliviero-gvb:
I'm facing the same bug. Do you plan to release this bug fix in the near future?

Delgan:
@rliviero-gvb Yes, a release is coming soon.

PR #1184, opened 2024-07-31
Fix for Runtime error when reinitializing logger

jeremyk:
We hit this error intermittently as we have to reinitialize our logger at least once during our application startup.

File "/opt/pysetup/.venv/lib/python3.11/site-packages/loguru/_logger.py", line 2031, in _log
    for handler in core.handlers.values():
RuntimeError: dictionary changed size during iteration

Fix is to copy the values to avoid the "changed size" error.

Fixes: https://github.com/Delgan/loguru/issues/1183

### DIFF
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -2033,7 +2033,7 @@ class Logger:
         for patcher in patchers:
             patcher(log_record)
 
-        # Create a copy to ensure safety when interating over handlers
+        # Create a copy to ensure safety when iterating over handlers
         for handler in list(core.handlers.values()):
             handler.emit(log_record, level_id, from_decorator, raw, colored_message)
 
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -2033,7 +2033,8 @@ class Logger:
         for patcher in patchers:
             patcher(log_record)
 
-        for handler in core.handlers.values():
+        # Create a copy to ensure safety when interating over handlers
+        for handler in list(core.handlers.values()):
             handler.emit(log_record, level_id, from_decorator, raw, colored_message)
 
     def trace(__self, __message, *args, **kwargs):  # noqa: N805


jeremyk:
Bump - hoping this can get a look sometime soon!

trim21:
this doesn't actually fix the problem, it only make less possible to happen

Delgan:
Thanks for investigating this bug and opening a PR, @jeremyk. Apologies for the delay.

Actually, the `logger` was supposed to be thread-safe in such scenario. Unfortunately, there is a mistake in the implementation.

The `_core._handlers` **must** be copied into a variable intended to be mutated locally.

This is done here for `logger.add()`: https://github.com/Delgan/loguru/blob/80f05fd2635a9f728565585fed02bed7c2be968a/loguru/_logger.py#L1006

And here for `logger.remove()`: https://github.com/Delgan/loguru/blob/80f05fd2635a9f728565585fed02bed7c2be968a/loguru/_logger.py#L1043

Once the `dict` containing the handlers has been updated as desired, the `_core._handlers` reference can be replaced with the previously copied one.

This allows the iteration through handlers for `logger._log()` to be lock-free, @trim21. It's kind of a copy-on-write mechanism. The `_log()` should acquire a reference to the `dict` that will never be modified.

Now, I said `logger.add()` and `logger.remove()` are actually copying the handlers as expected, but there is still a bug... This is because of that part: https://github.com/Delgan/loguru/blob/80f05fd2635a9f728565585fed02bed7c2be968a/loguru/_logger.py#L1053-L1059

The problem is that after the first iteration, the `self._core.handlers` is replaced with the locally copied `handlers` variable. But in subsequent iterations, this variable continues to be modified, while the value is now the same as in `_core.handlers`. This violates the previously stated principle that handlers `dict` can only be mutated locally, and that causes `RuntimeError: dictionary changed size during iteration` now that `_log()` and `remove()` are accessing the same `dict` in parallel.

The fix should be implemented in the `remove()` method instead of the `_log()` one. Do you want to update the PR @jeremyk or should I take care of that? It can be a simple as:

```patch
iff --git a/loguru/_logger.py b/loguru/_logger.py
index cc514b0..ffe4f80 100644
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -1040,7 +1040,7 @@ class Logger:
             )
 
         with self._core.lock:
-            handlers = self._core.handlers.copy()
+            handlers = self._core.handlers
 
             if handler_id is not None and handler_id not in handlers:
                 raise ValueError("There is no existing handler with id %d" % handler_id) from None
@@ -1051,6 +1051,7 @@ class Logger:
                 handler_ids = [handler_id]
 
             for handler_id in handler_ids:
+                handlers = handlers.copy()
                 handler = handlers.pop(handler_id)
 
                 # This needs to be done first in case "stop()" raises an exception
```

jeremyk:
I'm totally fine if you close this PR and take care of it. Thanks @Delgan !

Delgan:
Fixed by https://github.com/Delgan/loguru/commit/91e3646e8c0f1384d375bce49a5fb313dad5f7d6

Thanks again @jeremyk!

Issue #1201, opened 2024-09-11
very bad perf on `_datetime.datetime.__format__`


trim21:
current text format use `YYYY-MM-DD HH:mm:ss.SSS` and handled by `_datetime.datetime.__format__`, which is very slow comparing to stdlib version `%Y-%m-%d %H:%M:%S.%f%z`.

For example, I use `os.devnull` as sink and try these 2 format, `YYYY-MM-DD HH:mm:ss.SSS` will use 44s but `%Y-%m-%d %H:%M:%S.%f%z` use only 20s. (also it's 10s for stdlib logging)

```python
import os
import timeit

from loguru import logger as logger_new
from loguru_old import logger as logger_old

logger_new.info("hello world")
# exit(0)

logger_new.remove(0)
logger_new.add(os.devnull)
logger_old.remove(0)
logger_old.add(os.devnull)

def loguru_new_case():
    logger_new.info("hello {}", "b")


def loguru_old_case():
    logger_old.info("hello {}", "b")


print(timeit.timeit(loguru_new_case))
print(timeit.timeit(loguru_old_case))
```

Delgan:
Fixed by 80f05fd2635a9f728565585fed02bed7c2be968a.

Thank you for investigating possible performance improvements. :+1:

PR #1202, opened 2024-09-11
improve datetime perf

trim21:
benchmark with simple `logger.info("hello {}", "b")` with `os.devnull` sink, from 44s to 20s

ref: https://github.com/Delgan/loguru/issues/1201

### DIFF
--- a/loguru/_datetime.py
+++ b/loguru/_datetime.py
@@ -1,5 +1,7 @@
+import functools
 import re
 from calendar import day_abbr, day_name, month_abbr, month_name
+from collections.abc import Callable
 from datetime import datetime as datetime_
 from datetime import timedelta, timezone
 from time import localtime, strftime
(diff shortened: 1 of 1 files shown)


trim21:
... why are we still running tests on 3.5 and why we convert `[YYYY]` to `YYYY` in datetime formatting?

trim21:
also， why aware_now use time module instead of using datetime.now.astimezone and combine it as datetime？

trim21:
I think we can also include a fast path for default datetime format

Delgan:
Hey @trim21.

Thank you for your work in improving Loguru's performance.
I never realized that `datetime` formatting was so costly. Good catch. Caching a pre-computed formatter is a simple and excellent idea to reduce the overhead.

I've done a few experiments and I think we can go even further. We do not need to repeatedly parse the tokens with the regex. We can build a formattable string associated with runtime formatting functions.

See the diff here: https://github.com/Delgan/loguru/commit/80f05fd2635a9f728565585fed02bed7c2be968a

Here are a `timeit` command using a non-default format:

```bash
$ python -m timeit -s "from loguru import logger;logger.remove();logger.add(lambda _: None, format='{time:YYYY-MM-DD HH:mm:ss.SSSSSS Z}');logger.info('Hello')" "for _ in range(10):logger.info('Hello')"
```

Results:

```bash
# master
500 loops, best of 5: 426 usec per loop

# trim21/datetime-perf
1000 loops, best of 5: 225 usec per loop

# Delgan/faster-datetime
2000 loops, best of 5: 160 usec per loop
```

It's arguably an almost negligible improvement on what you've already done, but it pushes optimization further without adding much complexity.

Do you mind if I merge this commit instead? Do you think it could be improved?

Issue #173, opened 2019-11-21
I would love to `set better_exceptions.MAX_LENGTH = None`


wviana:
I've found on ExceptionFormatter class that it sets a default to 128. I couln't get a clue about some way to set it.

Delgan:
Well, there is actually no way to modify this parameter, and I did not planned to expose it.

I recently added a snippet in the documentation about [custom exception formatting](https://loguru.readthedocs.io/en/latest/resources/recipes.html#customizing-the-formatting-of-exceptions). Basically, the following solution should allow you to manually configure max line length, using the original [`better_exceptions`](https://github.com/Qix-/better-exceptions) library:

```python
import sys
from loguru import logger
import better_exceptions
better_exceptions.MAX_LENGTH = None

def format(record):
    format_ = "{time} {message}\n"

    if record["exception"] is not None:
        record["extra"]["exc"] = better_exceptions.format_exception(*record["exception"])
        format_ += "{extra[exc]}\n"

    return format_

logger.remove()
logger.add(sys.stderr, format=format)

string = f"A very l{'o' * 1000}ng string..."

try:
    1 / string
except:
    logger.exception("Error")
```

It's a little bit cumbersome, so I may or may not extend the API to facilitate custom exception formatting in the future.

daniel347x:
How is it that something so useful and important - the backtrace/diagnostics that Loguru wraps so well from better_exceptions (and I didn't even know that's how Loguru does it) - only has a years-old thread with like 2 comments discussing how to gain full visibility on these extraordinary backtraces?

Isn't Python used by many many thousands of programmers who presumably don't mind the extra few seconds that it takes to scroll to the right on a very long line if that means it will save hours of debugging time because you can actually see all of the variables and their values in the stack trace, rather than having them cut off "for the convenience" of not having long lines go off the right side of Notepad?

lyf2000:
as in #1252 was said
```py
    for h in logger._core.handlers.values():
        h._exception_formatter._max_length = None
```

worked well

PR #1224, opened 2024-10-26
Add a 'length' argument to the 'add' function supporting "max_length"

daniel347x:
This was a trivial change but of massive value for me to be able to simply pass a flag to the `logger.add()` function controlling the maximum line length in stack trace diagnostics.

The limit of 128 characters is too small, and the only reason I use Loguru is for this functionality, so I'm surprised it's not in place already given how trivial a change it is.

I do not object to the few seconds it takes to scroll to the right to see all variable values - not cut off by three dots - if it means saving hours or days of debugging time.

Note: The code uses the same approach as the existing design pattern of tying the default to a user-settable environment variable. Additionally, the default of 128 remains the same.

I also noticed the convention of single-word arguments to `add`, so I chose "length" as the name of this argument.


Delgan:
Hi @daniel347x.

Thank you for your interest in contributing to Loguru.

Getting back at your comment at https://github.com/Delgan/loguru/issues/173#issuecomment-243940946, it still seems natural to define a limit to the maximum length of the values displayed in the traceback. Since any arbitrary object can end up being printed, this avoids flooding if a long text or verbose dictionary happens to be logged.
Displaying values is convenient for a quick overview in the event of an error during development, but it cannot replace the use of a debugger.

I agree it seems quite straightforward to simply add a new `length` argument to `logger.add()`. However, I'm not necessarily in favor of such an approach, as the `add()` API is already quite complex and I'd like to avoid adding too many new options.

This is why for more advanced cases like yours, it's still possible to fully adapt exception formatting using a custom `format` function, as demonstrated in the above mentioned ticket.
Given that this is not an very common use case, it seems to be an acceptable solution, which doesn't require any changes to Loguru.

That being said, I understand it can frustrating  not to be able to easily adjust the width limit.

Your point makes me realize that it might be interesting to deprecate existing `backtrace` and `diagnose` arguments in favor of a new argument dedicated to exception formatting (and which could also deal with max length). This is all the more relevant as there are other formatter (such as the `rich` one) that could benefit from easier interfacing with Loguru logger. It was actually already suggested in #86.

Rough idea:

```python
logger.add(sys.stderr, format="{time} {message}", traceback=lambda exc: logger.traceback(exc, width=999))
```

Nothing concrete yet. But I think there's merit in considering a more generic approach, rather than a new `length` argument.

If it's acceptable to you in the short term, however, I'd be happy to support a new environment variable to set the maximum length of the current traceback. I'd prefer something like `LOGURU_TRACEBACK_WIDTH`. What do you think?

Issue #1207, opened 2024-09-18
improve lazy support


trim21:
since we are using new style format, it's possible to control how value are formatted with `__format__`

current `lazy` option is not necessary, we can add a `LazyValue` type

```python
import logging
import sys
import time
from typing import Callable, TypeVar

from loguru import logger

logger.remove()
logger.add(sys.stderr, level='INFO')

T = TypeVar('T')


class LazyValue:
    __slots__ = ('fn',)

    def __init__(self, fn: Callable[[], T]):
        self.fn = fn

    def __format__(self, format_spec: str):
        return format(self.fn(), format_spec)


def expensive_func() -> int:
    print('expensive_func called')
    time.sleep(1)
    return 1


logger.debug("hello {} from {}", LazyValue(expensive_func), "Alice")
logger.info("hello {} from {}", LazyValue(expensive_func), "Alice")
```

reneleonhardt:
Nice! Wouldn't it be easier to just accept a supplier function (i.e. `lambda: "Hello"` or an object method `None.__str__`) to be evaluated instead of a custom class?

trim21:
> Nice! Wouldn't it be easier to just accept a supplier function (i.e. `lambda: "Hello"` or an object method `None.__str__`) to be evaluated instead of a custom class?

I think that's very strange behavior for "formatting"

Delgan:
If anything, it already exists as an external library like [`lazy-string`](https://github.com/o3bvv/lazy-string) for example.

I agree it's more powerful than the current `opt(lazy=True)` design.

trim21:
> If anything, it already exists as an external library like [`lazy-string`](https://github.com/o3bvv/lazy-string) for example.
> 
> I agree it's more powerful than the current `opt(lazy=True)` design.

it only support old style percent formatting, not new style `{}` formatting

trim21:
now we have https://peps.python.org/pep-0750/ , any new thoughts?

Issue #782, opened 2023-01-13
A way to have lazy async function


Alcolo47:
I want to convert:
```
async def f():
    if logger.isEnabledFor(DEBUG):
        data = await heavy_computation()
        logger.debug(f"{data=}")
```
to loguru

async:  `logger.opt(lazy=True)` is not an option : #389

Delgan:
Thanks for the use case.

I have to admit that for now, I have no idea how to implement this given the current Loguru API based on `lazy` parameter. :(

xxmisiekxx:
wouldn't something like this work ?:

```
if logger.level <= loguru.levels.DEBUG:
    data = await heavy_computation()
    logger.debug(f{data=})
```

Delgan:
@xxmisiekxx Yes, that would work, but it comes with several drawbacks. First, it duplicates the role of the existing `lazy` argument for synchronous code. That makes the existence of `lazy` questionable, yet I prefer the `lazy` API since it avoids repetition of the level and reduces indentation. Secondly, it prevents potential optimizations, such as skipping the `heavy_computation()` if all handler filters return `False`.

Right now, the _best_ workaround I can think of is to split the logging call in two parts. The initial call is intended solely for "peeking" and assessing whether we should proceed further:

```python
async def main():
    should_log = False

    # The function will only be called if the message is to be logged.
    def peek():
        nonlocal should_log
        should_log = True

    logger.opt(lazy=True).debug("Preparing heavy computation", peek)

    if should_log:
        data = await heavy_computation()
        logger.debug("Heavy computation result: {}", data)

    logger.info("Done")
```

This can be abstracted into a re-usable helper function:

```python
async def log_if_needed(level, message, async_func):
    should_log = False

    def peek():
        nonlocal should_log
        should_log = True

    logger.opt(lazy=True).log(level, "Preparing asynchronous heavy computation...", peek)

    if not should_log:
        return

    result = await async_func()
    logger.log(level, message, result)

async def main():
    await log_if_needed("DEBUG", "Heavy computation result: {}", heavy_computation)
    logger.info("Done")
```

PR #1233, opened 2024-11-12
mark lazy mode is deprecated and add `LazyValue` for lazy format value

trim21:


trim21:
ci failed at codecov, not test/lint

Delgan:
Hi @trim21, thanks for opening a PR. I understand it refers to #1207.

I've been thinking about this lately, and to be honest I'm not very keen on the idea adding a new `LazyValue` class just to optimize lazy evaluations of message formatting. So far, the `logger` has been the only publicly importable component from `loguru`. That would be a significant change of this idiom for what seems to be a relatively small inconvenience.

I would welcome a new API for lazy evaluated arguments if it also address https://github.com/Delgan/loguru/issues/782. However, this is not necessarily easy.

Another thing to consider is [PEP 750](https://peps.python.org/pep-0750/). It's still draft and there's no guarantee that it will be accepted, but it could be interesting for our use case, without requiring a new API.

trim21:
also we have only sync api, I don't know there is a method to handle https://github.com/Delgan/loguru/issues/782

Delgan:
>  value in template str is still calculcated before logger function is called.

Indeed, but it's possible to pass a `lambda`: [Approaches to Lazy Evaluation](https://peps.python.org/pep-0750/#approaches-to-lazy-evaluation)

We could imagine than when a template string is used, then Loguru will convert callable arguments, but only if the log level requires it. This would make the `opt(lazy=true)` argument obsolete, while making it straightforward to combine with non-lazy arguments.

This PEP is actively discussed [here](https://discuss.python.org/t/pep750-template-strings-new-updates/71594).

> maybe we can add a api `logger.lazy(fn, *args, **kwargs)`?

I would prefer such API, yes. But still, that's a new method that increases the complexity of Loguru's API, for a marginal gain in my opinion.

> also we have only sync api, I don't know there is a method to handle https://github.com/Delgan/loguru/issues/782

Yes, unfortunately I don't know how to integrate it either.

trim21:
> We could imagine than when a template string is used, then Loguru will convert callable arguments, but only if the log level requires it. 

personal I think this is confusing and dangerous behaviour to call callable in arguments unless  explicit  specified with lazy or something

trim21:
> > value in template str is still calculcated before logger function is called.
> 
> Indeed, but it's possible to pass a `lambda`: [Approaches to Lazy Evaluation](https://peps.python.org/pep-0750/#approaches-to-lazy-evaluation)
> 
> We could imagine than when a template string is used, then Loguru will convert callable arguments, but only if the log level requires it. This would make the `opt(lazy=true)` argument obsolete, while making it straightforward to combine with non-lazy arguments.
> 
> This PEP is actively discussed [here](https://discuss.python.org/t/pep750-template-strings-new-updates/71594).
> 
> > maybe we can add a api `logger.lazy(fn, *args, **kwargs)`?
> 
> I would prefer such API, yes. But still, that's a new method that increases the complexity of Loguru's API, for a marginal gain in my opinion.
> 
> > also we have only sync api, I don't know there is a method to handle #782
> 
> Yes, unfortunately I don't know how to integrate it either.

what do you think about a new public API `loguru.utils.LazyValue` instead of `loguru.LazyValue`?

Delgan:
Thanks for the update.

Sorry, I should have catch that earlier but I realized there are two mains problems with such API:
* The expensive function will be called multiple time if the argument is to be formatted multiple times (e.g. `logger.info("{foo} {foo}", foo=logger.lazy(func))`).
* The `extra` dict which be populated with a `LazyValue` instead of the underlying value (for structured logging).

Even if we managed to get around these problems, to be honest I'm still not convinced by such an API change.

We're trying to fix a mere inconvenience. But there are other patterns that are not possible today and that it would be more interesting to handle. Notably #782 as said, or even:

```python
# There is no equivalent possible in Loguru.
if logger.isEnabledFor(logging.DEBUG):
    foo = expensive_func() 
    logger.info("Foo: %s", foo.summary)
    logger.debug("Foo details: %s", foo.details)
```

trim21:
these is easy to fix, we can just cache result

```python
class LazyValue:
    def __init__(self, fn, *args, **kwargs):
        self.fn = fn
        self.args = args
        self.kwargs = kwargs

    def __format__(self, format_spec: str):
        return format(self.__result, format_spec)

    def __str__(self):
        return str(self.__result)

    def __repr__(self):
        return repr(self.__result)

    @functools.cached_property
    def __result(self):
        return self.fn(*self.args, **self.kwargs)
```

Delgan:
Yes, I think a caching system is essential.

I was thinking again about this proposition the other day, and I believe we might be able to make it compatible with #782. Specifically, if the function to wrap is a coroutine, we could likely use [`create_task()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task) (as is already done for async sinks) to delay the logging call internally.

It’s possible that the object returned by `logger.lazy()` should be awaitable (similar to `logger.complete()`). I’m not certain yet, as I haven’t experimented with this approach.

Additionally, I recalled another improvement I’d like to implement: lazy arguments should not be evaluated if the `filter` of all handlers returns `False`. This seems important, given that there are use cases where the handler `level` is set to `0` and everything else is handled in a custom `filter` function.

With these two elements in mind, there’s plenty to consider regarding the final API, especially in terms of _when_ and _how_ to perform lazy evaluation. That said, I think it’s doable.

I know this doesn’t directly address your initial concerns in #1207. However, as I mentioned, I’m trying to get the best out of this of this future new method. Of course, this isn’t something that needs to be addressed in this PR (which could be merged as 1st step). These are just ideas at this stage.

In any case, it seems feasible to include these changes in the next minor release.

trim21:
> I was thinking again about this proposition the other day, and I believe we might be able to make it compatible with https://github.com/Delgan/loguru/issues/782. Specifically, if the function to wrap is a coroutine, we could likely use [create_task()](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task) (as is already done for async sinks) to delay the logging call internally.

I'm still not sure if this is possible, we can emit log to async sink with `asyncio.create_task` but it's not very doable with async LazyValue with sync sink.

trim21:
> Additionally, I recalled another improvement I’d like to implement: lazy arguments should not be evaluated if the filter of all handlers returns False. This seems important, given that there are use cases where the handler level is set to 0 and everything else is handled in a custom filter function.

This can be done with cache, maybe? we do not generate log message in `_log` but pass a class with `@functools.cached_property` and handlers will be able to get log line from a cached function.

and if no handle try to get log line, lazy value will not be called.

Delgan:
> This can be done with cache, maybe? we do not generate log message in `_log` but pass a class with `@functools.cached_property` with `def message() -> str: ...` and handlers will need to get log line from a cached function.

Yes, something along these lines. I was thinking maybe the `Lazy` could have a special `__getattribute__` that lazily evaluate the attribute access and cache the result. But that's yet to be defined.

trim21:
> > We can emit log to async sink with `asyncio.create_task` but it's not very doable with async LazyValue with sync sink, I think?
> 
> Here is roughly what I had in mind:


This doesn't look right to me...

for example, a logger with only `sys.stdout`/`sys.stderr`, which is a sync sink, right? 

a #782 example won't be able to write log message to it.

trim21:
OK, I got it. we will call _async_log method if any func of Lazy is coroutine function

trim21:
🤔 this will make internal implement very ugly because we will check every args and kwargs to see if it's Lazy value with coroutine

trim21:
for example

```python
class Lazy:
    def __init__(self, fn):
        self.fn = fn

    async def as_async_result(self):
        if asyncio.iscoroutinefunction(self.fn):
            return await self.fn()
        return self.fn()


def _sink(message: str):
    print(message)


async def _async_log(message: str, *args, **kwargs):
    _sink(await __get_real_log_line(message, *args, **kwargs))


def _sync_log(message: str, *args, **kwargs):
    _sink(message.format(*args, **kwargs))


async def __get_real_log_line(message: str, *args, **kwargs) -> str:
    # need to evaluate lazy value parallel here
    args = [(await v.as_async_result()) if isinstance(v, Lazy) else v for v in args]
    kwargs = {
        k: ((await v.as_async_result()) if isinstance(v, Lazy) else v) for k, v in kwargs.items()
    }
    return message.format(*args, **kwargs)


def is_async_lazy(*args, **kwargs):
    return any(asyncio.iscoroutinefunction(v.fn) for v in args if isinstance(v, Lazy)) or any(
        asyncio.iscoroutinefunction(v.fn) for v in kwargs.values() if isinstance(v, Lazy)
    )


def log(level: int, message: str, *args, **kwargs):
    if level < MIN_LEVEL:
        return

    if is_async_lazy(*args, **kwargs):
        task = asyncio.get_event_loop().create_task(_async_log(message, *args, **kwargs))
        tasks.append(task)
    else:
        _sync_log(message, *args, **kwargs)
```

trim21:
🤔 we will also need to maintain a backgroud task group in logger in-case they don't get freeed

trim21:
So I think we can do these in each seprated PR:

1. mark lazy mode deprecated and add `Lazy` for sync fn.
2. add asyncio support.
3. make lazy value evaluate after filter.

Delgan:
There are also other possible API that would not require to test `*args` and `**kwargs`, for example:

```python
logger.lazy(val=expensive_func).debug("My message: {val}")
```

But yes, once the API is fully defined, taking into account the technical aspect, it can be implemented in several PRs.

trim21:
Oh， I think this API looks like better

PR #856, opened 2023-04-27
Add timezone to default logging time

tim-x-y-z:
Hi thanks for the great library! 🤙

Started using it recently, and having resources in various locations I thought it would be great to have the timezone information in the default format. 
I realised that i could define a custom format for my need - but everything else just works out of the box, it felt that just this could be a nice default to have too.

Let me know your thoughts.

PR #1260, opened 2024-12-20
Fix tzinfo utcoffset error

bijlpieter:
Hey all, thank you for the wonderful library!

Since 0.7.3 I started running into tons of errors:
```
  File "xxx\loguru\_handler.py", line 184, in emit
    formatted = precomputed_format.format_map(formatter_record)
  File "xxx\loguru\_datetime.py", line 140, in __format__
    return _compile_format(fmt)(self)
           ~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "xxx\loguru\_datetime.py", line 23, in _loguru_datetime_formatter
    args = tuple(f(t, dt) for f in formatters)
  File "xxx\loguru\_datetime.py", line 23, in <genexpr>
    args = tuple(f(t, dt) for f in formatters)
                 ~^^^^^^^
  File "xxx\loguru\_datetime.py", line 106, in <lambda>
    "Z": ("%s", lambda t, dt: _format_timezone(dt.tzinfo or timezone.utc, sep=":")),
                              ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "xxx\loguru\_datetime.py", line 40, in _format_timezone
    offset = tzinfo.utcoffset(None).total_seconds()
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'total_seconds'
```

Like #856 is requesting, I'm "manually" adding a configurable timezone to my logging:
```
def tzconverter(record):
    record["time"] = record["time"].astimezone(tz=ZoneInfo("..."))

logger.configure(patcher=tzconverter)
```

The issue seems to be that the new `_format_timezone` calls `tzinfo.utcoffset` with `None`. And apparently `zoneinfo.ZoneInfo` objects return `None` for that input. I think the proper way to handle this would be to instead call `tzinfo.utcoffset` with the current time.

This PR implements that!

### DIFF
--- a/loguru/_datetime.py
+++ b/loguru/_datetime.py
@@ -36,8 +36,9 @@ def _default_datetime_formatter(dt):
     )
 
 
-def _format_timezone(tzinfo, *, sep):
-    offset = tzinfo.utcoffset(None).total_seconds()
+def _format_timezone(dt, *, sep):
+    tzinfo = dt.tzinfo or timezone.utc
+    offset = tzinfo.utcoffset(dt).total_seconds()
     sign = "+" if offset >= 0 else "-"
     (h, m), s = divmod(abs(offset // 60), 60), abs(offset) % 60
     z = "%s%02d%s%02d" % (sign, h, sep, m)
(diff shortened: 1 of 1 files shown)


bijlpieter:
Code coverage CI seems unrelated to my change... let me know if I should have another look

Delgan:
Oh, thanks a lot for fixing that!

Not sure why the implementation was changed from `.utcoffset(dt)` to `.utcoffset(None)` recently, it must have been an oversight. I'll add a few unit tests to avoid any regression in the future.

As a workaround for now, I think the `tzconverter()` patching function can be updated as follow:

```python
def tzconverter(record):
    zone = ZoneInfo("America/Los_Angeles")
    date = record["time"]
    seconds = zone.utcoffset(date).total_seconds()
    tzinfo = timezone(timedelta(seconds=seconds), zone.tzname(date))
    record["time"] = record["time"].astimezone(tz=tzinfo)
```

bijlpieter:
Fantastic Delgan, much appreciated, and happy coding :)

PR #1261, opened 2024-12-22
Package licence in built sdist

znichollscr:
The LICENSE is not packaged with the current build setup (check out PyPI, or also the failures here: https://github.com/conda-forge/loguru-feedstock/pull/39). I'm not totally familiar with flit, but this seems to work (although I'm not sure what best practice is exactly).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ dependencies = [
 description = "Python logging made (stupidly) simple"
 dynamic = ['version']
 keywords = ["loguru", "logging", "logger", "log"]
-license = { text = "MIT" }
+license = { file = "LICENSE" }
 name = "loguru"
 readme = 'README.md'
 requires-python = ">=3.5,<4.0"


znichollscr:
The created package distribution now includes the LICENSE (https://github.com/Delgan/loguru/actions/runs/12456650491/artifacts/2353755138).

In main, it is missing (https://github.com/Delgan/loguru/actions/runs/12222057044/artifacts/2290358305).

znichollscr:
@Delgan thanks for all your efforts with this project, it's super nice. Let me know if you want any other information or tweaks on this.

Delgan:
Thank you for identifying and fixing this, @znichollscr!

Ironically, this problem had already been addressed a few years ago in https://github.com/Delgan/loguru/pull/42.

Shortly after publishing 0.7.3 the other day (which is the 1st release based on Flit), I noticed the license file was unfortunately missing from the package. I wasn't sure it would cause downstream issues, since defining `{ text = "MIT" }` is supposedly equivalent. It appears definitely better to specify the file directly, then.

znichollscr:
Interesting. Maybe the conda ecosystem will update to recognize the licence
ID too soon. In the meantime, thanks

PR #1275, opened 2025-01-15
Make make_logging_logger() return log level to original state

kard3n:
When tests use the `make_logging_logger()` context with the name set to None, the log level of the default/root logger is set but not reset after the context is terminated. This causes some tests in tests/test_coroutine_sink.py that make calls to the logger in an asynchronous context fail if they are executed after. This is caused by asyncio adding a log (specifically `Using selector: EpollSelector`)  at `DEBUG` level when initialized, and due to the modified debug level this message is not discarded as expected.

For example, the following execution would fail: `pytest tests/test_interception.py tests/test_coroutine_sink.py`.

I fixed this by making `make_logging_logger()` remember the original log level, and restore it before exiting.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -315,7 +315,7 @@ def freeze_time(monkeypatch):
 
 @contextlib.contextmanager
 def make_logging_logger(name, handler, fmt="%(message)s", level="DEBUG"):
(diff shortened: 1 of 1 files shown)


Delgan:
Well done. I wouldn't have guessed it. Thanks a lot for catching and investigating this problem!

PR #1287, opened 2025-01-31
Add requirement for TERM not to be dumb to enable colorization

snosov1:
Hello!

This is a trivial change to disable coloring in case the terminal is dumb. Personally, I encountered the problem in Emacs terminal/compilation buffers - the coloring symbols are just printed as is (i.e. ASCII).

Technically, some more comprehensive logic can be implemented to derive color support from the TERM variable - but I'm not sure it's actually warranted. `toe /usr/share/terminfo` on my Ubuntu instance returns about 40 different terminal types - for some, it's clear whether the color is supported or not, for others - not that clear. 

Anyway, just opting out in case of `dumb` terminal seems like a sensible thing to do (and it solves my problem in Emacs).

Please, let me know what you think.

### DIFF
--- a/loguru/_colorama.py
+++ b/loguru/_colorama.py
@@ -33,7 +33,7 @@ def should_colorize(stream):
             return True
 
     try:
-        return stream.isatty()
+        return stream.isatty() and os.environ.get('TERM', '') not in ('dumb', )
     except Exception:
         return False
 


Delgan:
Hey @snosov1, thanks for your contribution! :+1: 

I agree with you that it seems fair to disable colors in this case. This is also what is done by [`supports-color.js`](https://github.com/chalk/supports-color/blob/ae809ecabd5965d0685e7fc121efe98c47ad8724/index.js#L106-L110) which I sometimes use as a reference, due to the completeness of their checks.

I'm merging the PR as-is but note that I'll probably move the check to [the code block right above](https://github.com/Delgan/loguru/blob/8653835b2fc1852930c2b224b190b60068634ed0/loguru/_colorama.py#L24-L33). This is because I think the special case for `$TERM="dumb"` should only apply when `stream` is the default standard output/error. If `stream` happens to be a custom `IOBase` provided by the user, the `$TERM` value should become irrelevant.

PR #1315, opened 2025-03-03
readme: Add a code snippet on how to use logprise

svaningelgem:
Added a little more context on how to use logprise.


Delgan:
Hi @svaningelgem.

Thanks for the suggestion, but I feel it falls outside the scope of the Loguru README. It should remain quite short, and I believe this snippet is better suited for the `logprise` README rather than Loguru's one.

svaningelgem:
I thought to add it as the snippet above also mentions apprise ;-)

PR #1337, opened 2025-04-24
chore: add Codecov test analytics

thomasrockhu-codecov:
Hi, Tom from Codecov here. I saw that you used Codecov already, and I thought you could benefit from catching any flaky or failed tests. We released a [Test Analytics](https://docs.codecov.com/docs/test-analytics) product that might be helpful and would love your feedback.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -60,6 +60,11 @@ jobs:
       run: |
         tox -e tests
       continue-on-error: ${{ matrix.allow-failure }}
+    - name: Upload test results to Codecov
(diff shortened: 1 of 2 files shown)


Delgan:
Hi. Thanks for opening a PR (which had become outdated by the time I reviewed it, sorry about that).

From what I understand about these new features, they may be useful for certain projects, but I don't believe Loguru would benefit significantly from them. There are no flaky tests, the tracebacks are already reported by Pytest, and the GitHub UI provides a sufficiently clear overview of the test results and coverage.

In short, I am satisfied with how Codecov integrates with the codebase currently, and see no need to extend it.

PR #1340, opened 2025-05-05
feat: add new Handler kwarg `multiprocessing_queue`

namoshizun:
This controls whether to enqueue messages to a multiprocessing-safe queue or a simple queue. If the user knows the program will never spawn child processes, settings this option to False may reduce the enqueue overhead.

BTW: for some reason the ruff linter insists on breaking up line 385 in __init__.pyi which misplaces the required "noqa: N805" rule on this line... I had to ignore the ruff rule in pre-commit for now until I figure out what went wrong 😑


loguru/__init__.pyi
```diff
@@ -377,7 +382,11 @@ class Logger:
     def exception(__self, __message: Any) -> None: ...  # noqa: N805
     @overload
     def log(
-        __self, __level: Union[int, str], __message: str, *args: Any, **kwargs: Any  # noqa: N805
+        __self,
```
injust, line 385: The `noqa: N805` comment should go on the `__self` line.

Issue #1344, opened 2025-05-14
Automatically add color to format strings


rawrke:
## Description 📝 

Add the ability to automatically color variables in format strings. This would automatically put color tags around curly braces within a log message, and then render it to a color. 

For instance, if the passed string to the logger is `"a: {}, b: {}, c: {}"`, then `loguru` will add a preconfigured color tag to the string before sending off to a sink, like so: `"a: <m>{}</m>, b: <m>{}</m>, c: <m>{}</m>"`. In this case, the tags `<m></m>` represent the color magenta.

## Example 📇 

**NOTE: The below example does not reflect the required configuration to enable this feature. A feature flag would likely be required to preserve backwards compatibility.**

```py
import sys
from loguru import logger
import math

logger.remove()

# Enable color rendering within a message
logger = logger.opt(colors=True)

# Add a sample logger
logger.add(
    sys.stdout,
    colorize=True,
    format="<g>[{time:YYYY-MM-DD HH:mm:ss}]</g><level>[{level}]</level> {message}",
    level="DEBUG"
)

temp = {"name": "Bob", "occupation": "Builder"}
logger.debug("Houston, we have a {}", "cat")
logger.debug("Mapping: {a}, π: {pi:.4f}", a=temp, pi=math.pi)
```

And the output, note that the contents within the curly braces `{}` are automatically colored and italicized. This would be customizable based on user desires.

<img width="1172" alt="Image" src="https://github.com/user-attachments/assets/84916189-3630-4704-9809-de78251bf300" />

rawrke:
Here is an example script that shows how to configure `loguru` to support this feature, with the changes implemented in #1345 .

```py
import sys
from loguru import logger
import math
import re
from collections.abc import MutableMapping
import datetime

logger.remove()

# _regex_pattern = re.compile(r"(\{.*?\})")
_regex_pattern = re.compile(r"(\{[^{}]*\})(?!\})")
_regex_repl = r"<lm><i>\1</i></lm>"

def insert_color_tags(record: MutableMapping):
    record.update(message=re.sub(_regex_pattern, _regex_repl, record["message"]))

# Enable color rendering within a message
logger = logger.patch(insert_color_tags)
logger = logger.opt(colors=True)

# Add a sample logger
logger.add(
    sys.stdout,
    colorize=True,
    format="<g>[{time:YYYY-MM-DD HH:mm:ss}]</g><level>[{level}]</level> {message}",
    level="DEBUG"
)

def main():
[...]

PR #1345, opened 2025-05-14
DRAFT: Insert colors around curly braces

rawrke:
Addresses https://github.com/Delgan/loguru/issues/1344

Still a WIP


rawrke:
This is ready for review. It addresses #1344. 

I wanted to add a feature that allows users to configure their logs to be colorful based on the format string that is passed in to the log message. 

I moved around when the patchers are applied to minimize the amount of code changes. Doing so allows users to configure their logs in more ways, an example is shown in the original issue (#1344)

PR #1352, opened 2025-06-18
Visualization of loguru's codebase

ivanmilevtues:
This PR introduces diagram visualization for loguru's codebase. You can see how the diagram renders here:
https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/loguru/on_boarding.md

The diagram documentation aims to ease the onboarding of new contributors. We believe that when starting at a new codebase one starts with a small component, but needs the big picture and the context of how that small component works within the big codebase. The diagram gives a good overview of what the main components are and how they interact with each other, then you can click on a component and get the same thing for that component. This way one can easily navigate to the related files of their first task and have the context of how they fit in the big picture. Let me know if this resonates with you!

We’ve also released a free GitHub Action that keeps the diagrams automatically updated as the code changes, so there's no ongoing maintenance burden.

Any feedback is more than welcome!

I'd usually open a discussion first, but you don't have them enabled for this repo so I decided to go ahead and open a PR.

Full transparency: we’re exploring this idea as a potential startup, but we’re still early and figuring out what’s actually useful to developers.

### DIFF
--- a/.codeboarding/Exception_Formatting_Utilities.md
+++ b/.codeboarding/Exception_Formatting_Utilities.md
@@ -1,0 +1,107 @@
+```mermaid
+
+graph LR
+
+    SyntaxHighlighter["SyntaxHighlighter"]
+
+    ExceptionFormatter["ExceptionFormatter"]
+
+    AnsiParser["AnsiParser"]
+
+    Colorizer["Colorizer"]
+
+    ExceptionFormatter -- "uses" --> SyntaxHighlighter
+
+    Colorizer -- "uses" --> AnsiParser
+
+```
+
+[![CodeBoarding](https://img.shields.io/badge/Generated%20by-CodeBoarding-9cf?style=flat-square)](https://github.com/CodeBoarding/GeneratedOnBoardings)[![Demo](https://img.shields.io/badge/Try%20our-Demo-blue?style=flat-square)](https://www.codeboarding.org/demo)[![Contact](https://img.shields.io/badge/Contact%20us%20-%20contact@codeboarding.org-lightgrey?style=flat-square)](mailto:contact@codeboarding.org)
+
+
+
+## Component Details
+
+
+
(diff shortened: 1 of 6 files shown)


Uzaaft:
It should be noted that Codeboarding uses LLMs for generating its outputs. 

And the PR author has just been going around and opening similar PRs like this all over the place.

ivanmilevtues:
Yes, the diagrams are generated with the help of Static Analysis and LLMs. The idea is that this way you can keep them up-to-date without having to manually update them.

Delgan:
Hi @ivanmilevtues. Thanks for considering Loguru for your demonstration.

In my opinion, documentation that lives outside the code tends to become outdated quickly, particularly when it includes diagrams or paragraphs that refer directly to the code components. The value it provides must significantly outweigh the burden of maintaining these docs for it to be worthwhile.

In a company setting, where teams are larger and new members join regularly, this type of onboarding material can be of great value. However, I don't believe that applies to a project like Loguru. The codebase is small, and new contributors are not very common. For that reason, I am not inclined to maintain and ensure the quality of additional documentation of this kind.

That said, I understand this documentation is generated and updated automatically. This can reduce the maintenance burden, but it introduces additional tooling complexity and dependencies. More importantly, any generated content would still need to be reviewed carefully. I could not blindly accept documentation written by an LLM.

Given that this is an automated process, it may be more appropriate to offer this onboarding documentation as an on-demand service. This would give newcomers the option to access an up-to-date view of the codebase when needed, and potentially explore it further interactively to build their understanding incrementally.

Hope this feedback can be somewhat useful to you.

ivanmilevtues:
Thanks for taking the time to look into this and answer @Delgan!

I am noting down all of it. Just wanted to touch on the point with maintaince burdon for reviewing the LLM Generated documentation. We are using an interative approach so only small portions of the docs will be update if the change requires such changes, so I believe we are covering on this end.

This said I understand your concerns on adding new dependency to the project. We will be open sourcing our logic to built trust with people that even if we dissapear (we are not going anywhere) they can continue using the service :)

ivanmilevtues:
Hey @Delgan and @Uzaaft wanted to do quickly update you, now we are open-source: https://github.com/CodeBoarding/CodeBoarding
There you can find exactly how we use the models and static analysis :)

PR #1368, opened 2025-08-05
adding docstring to recattrs sinks and parsers

aparna2198:
Added docstrings to three modules as part of improving documentation coverage for the repository.

### DIFF
--- a/loguru/_string_parsers.py
+++ b/loguru/_string_parsers.py
@@ -229,7 +229,7 @@ def parse_day(day: str) -> Optional[int]:
     if day.startswith("w") and day[1:].isdigit():
         day = int(day[1:])
         if not 0 <= day < 7:
-            raise ValueError(f"Invalid weekday value while parsing day: '%s'" % day)
+            raise ValueError(f"Invalid weekday value while parsing day: '%d'" % day)
     else:
         day = None
 
--- a/loguru/_string_parsers.py
+++ b/loguru/_string_parsers.py
@@ -114,7 +114,7 @@ def parse_size(size: str) -> Optional[float]:
     try:
         s = float(s)
     except ValueError as e:
-        raise ValueError("Invalid float value while parsing size: '%s'" % s)
+        raise ValueError("Invalid float value while parsing size: '%f'" % s)
 
     u = "kmgtpezy".index(u.lower()) + 1 if u else 0
     i = 1024 if i else 1000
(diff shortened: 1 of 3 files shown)


Delgan:
Hi.

Thank you for considering improving the docstrings. Loguru would definitely benefit from having more internal functions documented.

Please note that the CI is quite strict and will reject PRs that do not comply with the Ruff linter, in particular, n this case, the rules defined by pydocstyle: https://docs.astral.sh/ruff/rules/#pydocstyle-d
You can validate your changes locally by installing and running `pre-commit` in your development environment.

Also, be aware that Loguru aims to maintain backward compatibility with Python 3.5. I don't plan on changing that anytime soon, as I want to continue supporting it until the 1.0.0 release. For this reason, unfortunately, you cannot use modern features such as type hints or f-strings.

aparna2198:
@Delgan  I have fixed all the CI issues could you please mark this PR

loguru/_string_parsers.py
```diff
@@ -119,14 +230,29 @@ def parse_day(day):
     if day.startswith("w") and day[1:].isdigit():
         day = int(day[1:])
         if not 0 <= day < 7:
-            raise ValueError("Invalid weekday value while parsing day (expected [0-6]): '%d'" % day)
```
Delgan, line 122: Any reason to remove the range of accepted values from the error message?

loguru/_string_parsers.py
```diff
@@ -84,14 +169,25 @@ def parse_duration(duration):
         try:
             unit = next(u for r, u in units if re.fullmatch(r, unit, flags=re.I))
         except StopIteration:
-            raise ValueError("Invalid unit value while parsing duration: '%s'" % unit) from None
```
Delgan, line 87: I guess the surrounding `{}` braces can be removed.

loguru/_string_parsers.py
aparna2198: This was a miss, thanks for pointing. I'll add this back

Delgan:
Great, thanks for your work on the documentation and for keeping the strict CI happy.

Note that Loguru's public documentation follows the Numpy style. For consistency, I will likely update your docstrings to align with this convention rather than the Google style used in this PR. But this is very fine as it is, the most important thing being the addition of documentation.

aparna2198:
thanks @Delgan.

Issue #1200, opened 2024-09-11
add extra to default text output format


trim21:
This is a suggestion to change default loguru text output format.

There are many structured logging will include extra to text output as `key=value` format, hope loguru could do this, in someway, for example ` key1=value1 key2=value2`

golang stdlib log/slog:
```
2022/11/08 15:28:26 INFO hello count=3
```

zerolog(golang)
![image](https://github.com/user-attachments/assets/920352bf-632b-445f-9a0b-64da94b7cf33)

pino(nodejs)
![image](https://github.com/user-attachments/assets/605e7be8-e527-409a-bbcf-c93cc8b79a0d)

structlog(python)
![image](https://github.com/user-attachments/assets/2f34ea2c-f8e3-46ac-957b-6a667e81a833)

Delgan:
Thanks for the comparison with different frameworks. That's very helpful.

I'd already thought about it and I'm all for it. Structured logging is the way to go.

My plan is to add a new entry to the log record, for example a `data` dict. This would be filled by the keyword arguments of the logged message instead of `extra` currently. This is because it should be viewed as semantically different.

Then, add `{data}` to the current default format, which would be nicely formatted (if non-empty).

```python
logger.info("Hello world", foo="bar", baz=123)
# => 2024-09-22 19:39:58.381 | INFO     | __main__:<module>:1 - Hello world (foo="bar", baz=123)
```

trim21:
will kwargs from `logger.contextualize` be included?

Delgan:
Initially, I believed it should not, but maybe I'm wrong. I haven't given it enough thought. I need to look into the current usage made of `contextualize()` in relation to `bind()`.

elvizlai:
that should be really useful.

but for example :
logger.opt().bind(a=1).info("good {}, nice {b}", "morning", b="job", c="come on")

may be logout:
# => 2024-09-22 19:39:58.381 | INFO     | __main__:<module>:1 - good morning, nice job (c="come on")

should "a=1" included?

and I also read loguru source code, it can not clearly distinguish between 'bind' and 'format-placeholder', maybe we should add some new func with propagation to handler this.

boring but works like this

logger.opt().with(a=1, b="2").with(c=3).info("xxx")
# => - xxx (a=1, b="2", c=3)

Delgan:
@elvizlai Regarding the following example:

```python
logger.bind(a=1).info("good {}, nice {b}", "morning", b="job", c="come on")
```

[...]

PR #1386, opened 2025-09-29
WIP: Expose the default format as `DEFAULT_FORMAT`

GitToby:
I tend to have a quick solution to expose my logging with 
```python
_ = logger.remove()
_ = logger.add(sys.stderr, format=f"{_defaults.LOGURU_FORMAT} | {{extra}}")
```

However, accessing private modules is not ideal for code review or API stability. This will expose the format so I can extend it when specifying logging formats.

I'm also thinking this might be nice to have in the formatting step, so that we could define a message as

```python
_ = logger.remove()
_ = logger.add(sys.stderr, format="{defalt_format} | {extra}")
```

Thoughts welcome - could just be a me thing.

### DIFF
--- a/loguru/__init__.py
+++ b/loguru/__init__.py
@@ -10,10 +10,11 @@ import sys as _sys
 from . import _defaults
 from ._logger import Core as _Core
 from ._logger import Logger as _Logger
+from ._defaults import DEFAULT_FORMAT
 
 __version__ = "0.7.3"
 
-__all__ = ["logger"]
+__all__ = ["logger", "DEFAULT_FORMAT"]
 
 logger = _Logger(
     core=_Core(),
--- a/loguru/_defaults.py
+++ b/loguru/_defaults.py
@@ -28,13 +28,15 @@ def env(key, type_, default=None):
 
 
 LOGURU_AUTOINIT = env("LOGURU_AUTOINIT", bool, True)
-
+DEFAULT_FORMAT = (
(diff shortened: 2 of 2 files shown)


Delgan:
Hi.

Thanks for the suggestion but the default format isn't meant to be exposed publicly. Instead, I suggest copy-pasting it from source and simply updating it according to your preferences. If this proves too repetitive in the long run, you can also define the environment variable `LOGURU_FORMAT` once and for all.

Otherwise, you can hack your way around using the built-in [`inspect`](https://docs.python.org/3/library/inspect.html) module:

```python
default_format = inspect.signature(logger.add).parameters["format"].default
```

Note that I eventually plan to make `{extra}` field part of the default format

GitToby:
gotcha - it really is just to expose the `extras` in the default format.

its a super neat struct logging approach, which we have to expose every time I init a program. 

Is there an issue or something for the `{extra}` field to be part of the default format? I'd be happy to give it a crack and open a PR.

Delgan:
The following ticket could be considered related: https://github.com/Delgan/loguru/issues/1200.

But for now, there are only preliminary discussions, no precise specifications. It's actually more complicated than it looks like, because it's tied to the API change I'm planning for `*args` / `**kwargs` and the separation with `bind()` / `contextualize()`. All this to pivote toward better structured logging support. To be fair, there is still some side work to be done before a PR could be envisaged.

Issue #1411, opened 2025-10-30
Logging with level as integer does not map to expected log level name


RafaelWO:
Consider the following example:

```python
import logging

from loguru import logger

logger.debug("Debug with `logger.debug()`")
logger.log(logging.DEBUG, "Debug with `logger.log(logging.DEBUG)`")
logger.log("DEBUG", "Debug with `logger.log('DEBUG')`")
```

If I run this with `loguru==0.7.3`, I get the following output:
```
2025-10-30 09:34:32.031 | DEBUG    | __main__:<module>:5 - Debug with `logger.debug()`
2025-10-30 09:34:32.031 | Level 10 | __main__:<module>:6 - Debug with `logger.log(logging.DEBUG)`
2025-10-30 09:34:32.031 | DEBUG    | __main__:<module>:7 - Debug with `logger.log('DEBUG')`
```

But I expected:
```
2025-10-30 09:34:32.031 | DEBUG    | __main__:<module>:5 - Debug with `logger.debug()`
2025-10-30 09:34:32.031 | DEBUG    | __main__:<module>:6 - Debug with `logger.log(logging.DEBUG)`
2025-10-30 09:34:32.031 | DEBUG    | __main__:<module>:7 - Debug with `logger.log('DEBUG')`
```

Since the `logger.log()` method is annotated to accept `int | string`, I would have expected the level name to be `DEBUG` if the level number 10 for debug is supplied.

Is that expected behaviour?

I am happy to provide a fix via a PR for this 🙂

Delgan:
Hi @RafaelWO.

Thanks for the report. However, this is actually the expected behavior. Unlike the standard library, Loguru levels are identified exclusively by their name (and not by their severity). Multiple level names can be assigned to the same level severity, therefore levels severity are not mapped to any level name since doing so would possibly face ambiguities. If an integer is provided to `log()`, the level is interpreted as an anonymous one.

Just for completeness, here are a few of related tickets:
- https://github.com/Delgan/loguru/issues/209
- https://github.com/Delgan/loguru/issues/490
- https://github.com/Delgan/loguru/issues/599
- https://github.com/Delgan/loguru/pull/1106
- https://github.com/Delgan/loguru/issues/1011
- https://github.com/Delgan/loguru/issues/1054

Well, this is arguably more than "a few"... Given how frequently this comment comes up, I should definitely add this clarification to the Loguru documentation.

RafaelWO:
[...]

PR #1412, opened 2025-11-03
Document anonymous log level for `log()`

RafaelWO:
This change adds the info about how integer log levels are interpreted
when using `logger.log(10, ...)` and link to the related FAQ entry.

Related to #1411


Delgan:
Thanks, that is indeed a worthwhile addition!

PR #1429, opened 2026-01-14
Trivial grammar fix

aura-ix:
See diff.


aura-ix:
oops, will fix. didn't consider that tests relied on this.

Delgan:
Kind of ironic... Thank you!

Issue #1444, opened 2026-03-11
Import multiprocessing only on enqueue=True


sebastian-goeldi:
Hi,

Love loguru :)

Would it be possible to make the import of multiprocessing conditional on using `enqueue=True`?

Reasoning is, I'd like to use this in a project where i depend on loguru but want to compile it to pyodide/wasm. pyodide does not currently support multiprocessing (due to browser limitation) and throws an error on import of multiprocessing.

To my knowledge this does not hurt performance significantly anymore.

EDIT:

I made a patch with claude that works (and supposedly passes the loguru tests, haven't verified anything yet). I can submit the PR if desired, but I am not sure how good it is.

Delgan:
Hi @sebastian-goeldi. :)

Surely, we can try to make Loguru compatible with pyodide.

However, I am not familiar with this tool. Is pyodide incompatible with any library importing the `multiprocessing` module? Because, then, I imagine that many external libraries would be rendered unusable.

[The official documentation](https://pyodide.org/en/stable/usage/wasm-constraints.html) seems to imply that `multiprocessing` can be imported but not used:

> The following modules can be imported, but are not functional due to the limitations of the WebAssembly VM:
> - multiprocessing
> - threading
> - sockets

Could you please elaborate on the error you are experiencing? If a mere [`import multiprocessing`](https://github.com/Delgan/loguru/blob/8c7d33c4cd349f5b03992973a6b564b01a055359/loguru/_handler.py#L3) causes an `Exception`, that would seem to contradicts the expectations.

> I made a patch with claude that works (and supposedly passes the loguru tests, haven't verified anything yet). I can submit the PR if desired, but I am not sure how good it is.

Feel free to submit it in accordance with [the AI policy](https://github.com/Delgan/loguru/blob/8c7d33c4cd349f5b03992973a6b564b01a055359/CONTRIBUTING.rst#automated-contributions-policy) of this project. That is, only if you're confident you understand the patch well and, ideally, can provide unit tests proving it works correctly. Generating code with an LLM is cheap, I can do it on my end. The benefical added value comes from the time spent reviewing, curating and testing the changes. ;)

But I'd like to make sure we are fixing the right problem, first.

sebastian-goeldi:
[...]

PR #1446, opened 2026-03-13
Provide fallback for multiprocessing if it's not properly available

sebastian-goeldi:
This PR addresses https://github.com/Delgan/loguru/issues/1444

Disclaimer: This was coded by AI (Claude Code with one of the Opus)
The changes are small enough that I confidently can say I understand the code changes and they seem reasonable to me.

Changes:
The biggest difference is using `os.getpid()`  instead of the multiprocessing `current_process().ident`. I think this should be equivalent, but I wanted to point this one out as  I am not super deep into multi processing and threading.

As mentioned in the issue, this is at the moment mainly targeted at [pyodide](https://pyodide.org). pyodide provides a way for python to run completely in the browser through webassembly. A minimal working example is the following (this requires to build the wheel of loguru, I built it with `python -m build`)

```html
<!doctype html>
<html>
  <head>
      <script src="https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.js"></script>
  </head>
  <body>
    Pyodide test page <br>
    Open your browser console to see Pyodide output

    <img id="png-output" width="800" height="600" style="display:none;" />
    <script type="text/javascript">
      function displayPngBase64(b64) {
        const img = document.getElementById("png-output");
        img.src = "data:image/png;base64," + b64;
        console.log(img.src)
        img.width = 800;
        img.height = 600;
        img.style.display = "block";
      }
    </script>
    <script type="text/javascript">
      async function main(){
        let pyodide = await loadPyodide();
        await pyodide.loadPackage("micropip");
        console.log(await pyodide.runPythonAsync(`
            import micropip

            await micropip.install("./loguru-0.7.3-py3-none-any.whl")

            import loguru
            import sys
            
            loguru.logger.add(sys.stdout)
            loguru.logger.warning("hello from loguru")
        
        `));
      }
      main();
    </script>
  </body>
</html>
```

This gets printed to the console of the browser after opening it in the browser (use `python -m http.server 8000` to serve it and go to `http://127.0.0.01:8000` to access it) and opening the developer interface to see console and its log output

image

TODO:

- Add tests. Maybe @pepijndevos could help out here. I am not very familiar with emscripten & pyodide, but I know pyodide provides a way to write tests for it
- I think it currently still can fail if the context is a string? I don't know yet how to change the context to that :P

### DIFF
--- a/loguru/_handler.py
+++ b/loguru/_handler.py
@@ -1,6 +1,5 @@
 import functools
 import json
-import multiprocessing
 import os
 import threading
 from contextlib import contextmanager
@@ -89,6 +88,8 @@ class Handler:
                 self._decolorized_format = self._formatter.strip()
 
         if self._enqueue:
+            import multiprocessing
+
             if self._multiprocessing_context is None:
                 self._queue = multiprocessing.SimpleQueue()
                 self._confirmation_event = multiprocessing.Event()
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -100,14 +100,13 @@ import builtins
 import contextlib
 import functools
 import logging
+import os
 import re
 import sys
 import threading
 import warnings
 from collections import namedtuple
 from inspect import isclass, iscoroutinefunction, isgeneratorfunction
-from multiprocessing import current_process, get_context
-from multiprocessing.context import BaseContext
 from os.path import basename, splitext
 from threading import current_thread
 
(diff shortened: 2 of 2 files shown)


sebastian-goeldi:
Not necessary anymore since I am an idiot and didn't notice that it actually works after disabling all multiprocessing and enqueue

Issue #17, opened 2018-12-14
Data scrubbing


360developer:
*sentry.io* (https://docs.sentry.io/data-management/sensitive-data/) comes with a nice way to scrub sensitive data (as stipulated by data privacy guidelines or regulations, such as the GDPR).

We are looking for a convenient way to hide/obfuscate sensitive data which can be enabled/disabled by configuration for staging/production environments.
We would like to use a predefined filter (identical to sentry.io) that allows customization, so we could add more filters to it.

Delgan:
Hi @360developer, does a `filter` function would suit your needs?

However, the filtering process is not as easy that with Sentry.io, because logging mainly uses string messages.

This entirely depends on how you intend to log the sensitive data.

For example, if you planned to do something like:

```python
logger.info("It's ok, no sensitive data")
# Add "username" and "password" to the "extra" dict of each recorded log
user_logger = logger.bind(username="user", password="xyz) 
user_logger.info("Warning, sensitive data bound")
```

The filter function of your handler would look like:

```python
def scrub(record):
    forbidden = ["username", "password", "apikey"]
    for field in forbidden:
        if field in record["extra"]:
            return False
    return True

logger.add("file.log", filter=scrub)
```

Alternatively, you could decide which logs should be filtered-out while calling the logger function:

```python
logger.info("It's ok")
logger.bind(sensitive=True).info("Username is '{}' and password is '{}'", "user", "xyz")
```

And use a filter function like this:

```python
def scrub(record):
    if "sensitive" in record["extra"]:
        return False
    return True
```

I think I may need more examples about how you plan to use `loguru` to conditionally log sensitive data.

Delgan:
If you need more support, feel free to re-open or comment this issue further. 👍

360developer:
@Delgan - thanks, your comment helped. We have to better structure the way we create the log entries. Then a filter does the job.

fredrikaverpil:
@Delgan what about if I have something today for the standard library logging module:

```python
import logging
import re


class SensitiveFormatter(logging.Formatter):
    """Formatter that removes sensitive information in urls."""
    @staticmethod
    def _filter(s):
        return re.sub(r':\/\/(.*?)\@', r'://', s)

    def format(self, record):
        original = logging.Formatter.format(self, record)
        return self._filter(original)
```

Is it possible to mask parts of the log string like this using loguru?

fredrikaverpil:
[...]

PR #1447, opened 2026-03-18
Add support for filtering out secrets when `diagnose=True`

PerchunPak:
Useful if you still want to have more context in the production logs, but don't want to leak the passwords or other secrets. Usage example:

```python
logger.add("out.log", diagnose=True, diagnose_excludes=["myS3cr3tP@ss!"])
```

Any value added to `diagnose_excludes` is replaced with `<hidden>` wherever it appears in the exception output, including variable values and the exception message.

```py

  File "tests/exceptions/source/diagnose/excludes.py", line 23, in <module>
    connect_to_db(password)
    │             └ '<hidden>'
    └ <function connect_to_db at 0xDEADBEEF>

  File "tests/exceptions/source/diagnose/excludes.py", line 18, in connect_to_db
    raise TimeoutError("tried to connect to " + repr(connection_string))
                                                     └ 'foo bar <hidden> baz'

TimeoutError: tried to connect to 'foo bar <hidden> baz'
```


Delgan:
Hi.

Log Scrubbing is an important topic, but I'm not convinced this is the right approach. Declaring secrets statically like this appears quite brittle, and it covers too few cases to be truly useful in production. This is a complex subject which I don't think can be handled by Loguru itself. I would rather recommend wrapping your secrets in a strong type that disallow representation, or using solutions specifically designed for this purpose, such as Datadog or similar tools.

PerchunPak:
My personal use case is that SQLAlchemy prints the connection URL when trying to connect to an offline database. Datadog is totally overkill for our needs.

I also don't like to declare secrets globally, but I don't know how to do it otherwise, since Python does not have truly private fields. And if something has gained access to the runtime, I think the user has bigger problems (e.g., access to the filesystem and env variables)
https://www.google.com/search?q=python+litellm+hack

Delgan:
You can possibly implement you scrubbing mechanism within a custom sink, as such:

```python
from loguru import logger
import sys

password = "myS3cr3tP@ss!"


def scrubbing_sink(message):
    diagnose_excludes = [password]
    for secret in diagnose_excludes:
        message = message.replace(secret, "<hidden>")
    print(message, file=sys.stderr)


logger.remove()
logger.add(scrubbing_sink, diagnose=True, colorize=True)


def connect_to_db(password):
    raise ValueError()


try:
    connect_to_db(password)
except Exception as e:
    logger.exception("Failed to connect to database")
```

It's still a bit fragile, in my opinion.

Others alternatives might be listed in https://github.com/Delgan/loguru/issues/17.

PerchunPak:
Oh, that seems nice, will play with it on Monday. Though, how can this break? My implementation seems to handle `repr` fine

Delgan:
> Though, how can this break?

The first problem is that the filter runs after the exception has been processed and formatted (in particular, after adding colors). The filter only works based on the (somewhat fragile) assumption that Loguru will not modify the raw string.

For Loguru, it works because ANSI escape sequences (for colors) are added around the entire string: `'myS3cr3tP@ss!'` => `\033[1;36m'myS3cr3tP@ss!'\033[0m` => `\033[1;36m'<hidden>'\033[0m`. However, if the implementation were to change, your filter could be compromised. For example, the `rich` library also add specific colors to integer inside strings, and in such a case the inner ANSI characters would prevent the string to match the filter.

Another very concrete example is that Loguru truncates values longer than 128 characters while formatting variables. You can replace `password = "myS3cr3tP@ss!" + "=" * 128` in the previous snippet, and you'll observe that the password get leaked in the output.

That is why this "post filter" approach is not robust and cannot be recommended.

Pydantic offers interesting strong types as I discussed previously, such as [`SecretStr`](https://docs.pydantic.dev/2.12/api/types/#pydantic.types.SecretStr). Escaping occurs automatically when the object itself is requested to be rendered, the secret string is never formatted, therefore it is not subject to the limitations described above. The downside is that it does not compose very well with other non-secret strings, and you need to wrap all strings to `SecretStr` manually.

PerchunPak:
Sorry, I am no longer in the company for which I made this PR, so lost interest in it for a while. But I still want to see this in loguru

> The first problem is that the filter runs after the exception has been processed and formatted (in particular, after adding colors).

It runs right before colorization happens
https://github.com/Delgan/loguru/pull/1447/changes#diff-ccb032e0ec345ddcc0a87baaacf7236552af45a80f1df3770215bb1d190776beR489

> Loguru truncates values longer than 128 characters while formatting variables

The same here, this filter runs right before truncation happens
https://github.com/Delgan/loguru/pull/1447/changes#diff-ccb032e0ec345ddcc0a87baaacf7236552af45a80f1df3770215bb1d190776beR357

PR #1455, opened 2026-04-01
fix(logger): Corrects callable formatter handling of automatic line ending and exception suffix (#1454)

### LINKED ISSUE #1454: Inconsistency between format string and dynamic formatting
Thanks for this library - seems very promising, as far as I can say. Just trying it out to see if it fits my needs (release v0.7.3).

I tried to print some extra-Infos which aren't contained in the default format. 
Thanks to your reply here (#1164), I could extend the default format (which - as it's internal - shouldn't be done this way for production, I know) as follows:
```
logger.add(sys.stderr,
    format=lambda rec: loguru._defaults.LOGURU_FORMAT
                       + (" @ {extra}" if rec['extra'] else "")
                       + "\n{exception}",         ###  <-- INCONSISTENCY
)
```

At first, I hadn't added the last row containing line break and exception, which caused the logging to miss line breaks and exception bodies.

I dug deeper into this, why the default format - which doesn't contain exception and linebreak - works differently, just comparing those two loggers one after the other for their format and behaviour at exceptions:

```
logger.add(sys.stderr, format="{message}")
logger.add(sys.stderr, format=lambda r:"{message}")
```

Even if the format string is effectively the same simple string (`"{message}"`), the lambda-generated string results in a different format than the plain string.

I think, this is the relevant part in the code, where a string format gets extended by terminator and exception (L989), but the dynamic formatter does not (L1002).

https://github.com/Delgan/loguru/blob/5ec8d0034ef38f9978b023b5d939f437538b49f5/loguru/_logger.py#L987-L1003

For me, this distinction seems counterintuitive and as far as I could find, is not [documented](https://loguru.readthedocs.io/en/stable/api/logger.html#loguru._logger.Logger.add).

> format ([str](https://docs.python.org/3/library/stdtypes.html#str) or [callable](https://docs.python.org/3/library/functions.html#callable), optional) – The template used to format logged messages before being sent to the sink. 

I'd suggest to unify the behaviour between string and dynamic formatting, or at least document the differences. 
(And ideally, to add the extra-output in the default format.)

RedZapdos123:
### Description:
I noticed that this bug happened because string formatters were internally extended with the sink terminator and `{exception}`, while callable formatters were used as is. This made two equivalent templates behave differently.

This PR addresses it by applying the same suffix behavior to callable formatters. It passes the sink terminator into the handler and appends terminator + {exception} for dynamic templates before rendering. I also updated docs and tests to reflect the unified behavior.

Closes #1454

### AI Assistance Disclosure:
- **Tool(s) used:** GitHub Copilot GPT-5.3-Codex.
- **Scope:** Used AI to implement targeted changes in logger/handler behavior, update tests, and update documentation/changelog.
- **Verification:** I reviewed all changed lines and validated with local runs in WSL using `tox -e tests` and `tox -e lint`.

### Checklist:
- [x] I have reviewed all changes in this PR myself.
- [x] I have added or updated relevant test cases for this fix.
- [x] I have run linting and formatting checks.

Screenshot 2026-04-01 141456

Attached a screenshot of the tests for validation.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@
 - Make ``logger.catch()`` usable as an asynchronous context manager (`#1084 <https://github.com/Delgan/loguru/issues/1084>`_).
 - Make ``logger.catch()`` compatible with asynchronous generators (`#1302 <https://github.com/Delgan/loguru/issues/1302>`_).
 - Improve feedback for invalid format keys in logger format strings (`#1450 <https://github.com/Delgan/loguru/issues/1450>`_, thanks `@Krishnachaitanyakc <https://github.com/Krishnachaitanyakc>`_).
+- Make callable log ``format`` append the same line ending and ``{exception}`` suffix as string formats (`#1454 <https://github.com/Delgan/loguru/issues/1454>`_).
 
 `0.7.3`_ (2024-12-06)
 =====================
--- a/docs/resources/troubleshooting.rst
+++ b/docs/resources/troubleshooting.rst
@@ -279,7 +279,7 @@ Instead, the function should return a string with placeholders, like this::
 
     def correct_dynamic_format(record):
         # This is correct because it returns a template string.
-        return "{time} - {level} - {message}\n{exception}"
+        return "{time} - {level} - {message}"
 
     logger.add(sys.stderr, format=correct_dynamic_format)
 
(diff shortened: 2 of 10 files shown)


Delgan:
The API cannot be changed this way. First, this is a breaking change, and it needs a deprecation period. Secondly, unconditionally adding `\n{exception}` would prevent legitimate use cases such as JSONL logging.

PR #1465, opened 2026-04-06
make mypy truly optional for running the tests

mr-c:


mr-c:
With this PR, tests will pass even if none of the mypy related packages are installed (`mypy`, `pytest-mypy-plugins`), and without needing `--ignore=tests/typesafety/test_logger.yml` being passed to `pytest`.

Delgan:
Thanks for the PR.

I'd expect `mypy` to be a required dev dependency of Loguru, just like the others (namely `freezegun` and `colorama`). That being said, I can understand giving a static analyzer special treatment and considering it optional. The `test_type_hinting.py` file only exists because `pytest-mypy-plugins` couldn't run on older Python versions, and the `test_logger.yml` files are already automatically and silently skipped when the plugin isn't installed anyway. If making Mypy optional helps package maintainers, then it makes sense to go that route.

mr-c:
Thank you @Delgan ; yes, this helps us not to not break the Debian package of loguru just because a new (incompatible) version of `mypy` was released.

PR #1474, opened 2026-05-21
Fix parsing of 12-hour rotation times without seconds

c-tonneslan:
`parse_time` can't parse a 12-hour rotation time that has hours and minutes but no seconds:

```python
>>> from loguru._string_parsers import parse_time
>>> parse_time("1:30 pm")
ValueError: Unrecognized format while parsing time: '1:30 pm'
>>> parse_time("11:45 AM")
ValueError: Unrecognized format while parsing time: '11:45 AM'
```

The seconds-bearing forms work fine (`parse_time("1:30:00 pm")` → `13:30:00`), which points at the format list. It has:

```python
"%I %p",
"%I:%M %S",      # <- meant to be "%I:%M %p"
"%I:%M:%S %p",
"%I:%M:%S.%f %p",
```

Every other `%I` (12-hour) entry pairs the time with the `%p` AM/PM marker. `"%I:%M %S"` instead expects a bare seconds field after a space — a format no real input produces — so `"1:30 pm"` matches none of the patterns even though the regex gate above accepts it.

Changed it to `"%I:%M %p"`. Added `("1:30 PM", ...)` to the `test_time_rotation` parametrize list (it raises `ValueError` on `master` and passes with the fix); the full `test_filesink_rotation.py` suite still passes.


Delgan:
Please, [disclose usage of AI and agents](https://github.com/Delgan/loguru/blob/master/CONTRIBUTING.rst#automated-contributions-policy) next time. Otherwise, good fix, thanks.

c-tonneslan:
Understood, will disclose on any future PRs to loguru. Thanks for taking this one.

PR #1476, opened 2026-05-22
remove unnecessary python upper bound

dimbleby:
There's more or less consensus these days that this upper bound in python-requires is undesirable, certainly for pure python projects.

The upper bound was added at #1140 migration to pyproject.toml


Delgan:
Yeah, I re-read recommendations about `requires-python` and I think you're right. Thanks!

PR #1477, opened 2026-05-24
Fix parse_size docstring: it returns bytes, not bits

c-tonneslan:
The function's docstring claims it returns bits in two places, but
the implementation actually returns bytes:

- a bare \`b\` suffix (the bit-literal case) is divided by 8
- a \`B\` suffix passes through unchanged
- \`rotation_size\` in \`_file_sink.py\` compares the result against
  \`file.tell() + len(message)\`, which is bytes

So \`parse_size("10MB")\` returns 10,000,000, which is bytes. Fixed
the wording in both places.

Disclosure: AI assisted with reading through the file to find this.

### DIFF
--- a/loguru/_string_parsers.py
+++ b/loguru/_string_parsers.py
@@ -100,7 +100,7 @@ class Frequencies:
 
 
 def parse_size(size: str) -> Optional[float]:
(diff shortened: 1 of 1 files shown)


Delgan:
Thanks.

PR #1483, opened 2026-06-28
Fix Record extra type annotation from Dict[Any, Any] to Dict[str, Any]

### LINKED ISSUE #1482: Record type bug
```python
from loguru import Record

def func(record: Record) -> None:
    print(record)  # using in any function, forcing mypy to do typecheck
```
```
mypy: Expression type contains "Any" (has type "Record")
```

its because of 
```python
class Record(TypedDict):
    elapsed: timedelta
    exception: Optional[RecordException]
    # \/\/\/\/\/\/\/\/\/\/
    extra: Dict[Any, Any]
    # /\/\/\/\/\/\/\/\/\/\ 
    file: RecordFile
    function: str
    level: RecordLevel
    line: int
    message: str
    module: str
    name: Optional[str]
    process: RecordProcess
    thread: RecordThread
    time: datetime
```

maybe separate Record and Record with context or something like that?

albin-george-kurian:
Fixes #1482

## What changed
- `Record["extra"]` retyped from `Dict[Any, Any]` to `Dict[str, Any]`
- `Logger.configure(extra=...)` retyped the same way
- Added `tests/typesafety/test_record_extra.yml` with 5 test cases

## Why
`bind(**kwargs)` only accepts string keyword arguments, so `Any` keys
were inaccurate. The `Any` key type caused mypy's `--disallow-any-generics`
to flag any function annotated with `Record` as tainted.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,7 +19,7 @@
 - Make ``logger.catch()`` usable as an asynchronous context manager (`#1084 <https://github.com/Delgan/loguru/issues/1084>`_).
 - Make ``logger.catch()`` compatible with asynchronous generators (`#1302 <https://github.com/Delgan/loguru/issues/1302>`_).
 - Improve feedback for invalid format keys in logger format strings (`#1450 <https://github.com/Delgan/loguru/issues/1450>`_, thanks `@Krishnachaitanyakc <https://github.com/Krishnachaitanyakc>`_).
-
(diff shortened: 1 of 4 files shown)


Delgan:
This does not fix anything.

```python
# demo.py
from loguru import Record

def func(record: Record) -> None:
    print(record)
```
```bash
$ mypy --disallow-any-expr demo.py 
demo.py:4: error: Expression type contains "Any" (has type "Record")  [misc]
Found 1 error in 1 file (checked 1 source file)
```

The `disallow_any_generics = True` added as unit test is irrelevant. We want to test `disallow_any_expr` as reported by #1482.

Please, [verify the implementation before submitting AI output](https://github.com/Delgan/loguru/blob/master/CONTRIBUTING.rst#automated-contributions-policy).

PR #1484, opened 2026-06-30
Fix first rotation of "<weekday> at <time>" ignoring the weekday

binggao1230:
### The bug

A `rotation="<weekday> at <time>"` rule fires its **first** rotation on the file's creation day instead of the requested weekday, when the sink is created on a different weekday at a time of day earlier than the target time.

```python
import datetime
from loguru import logger

# Sink created on a Sunday at 12:00, rotating every Tuesday at 13:00.
logger.add("app_{time}.log", rotation="tuesday at 13:00")
```

Started on Sunday 12:00, a spurious rotation fires that same **Sunday at 13:00** (producing an empty rotated file and starting a new one), before settling onto Tuesdays. Expected: nothing should rotate until **Tuesday 13:00**.

### Cause

In `Rotation.RotationTime.__call__`, the initial `limit` is set to the creation date at the target time of day. The weekday is carried only by `step_forward` (`forward_weekday`), which is applied **only when `limit <= start_time`**. When the target time of day is later than the creation time, `limit > start_time`, so the step is skipped and the limit stays on the creation day — the wrong weekday.

It only manifests when the creation weekday differs from the target weekday **and** the target time of day is later than the creation time, which no existing test exercised (they all start Sunday 12:00 with same-weekday or earlier-time rules).

### The fix

Track the requested weekday on `RotationTime` and also step forward when the initial limit lands on the wrong weekday, so the first rotation happens on the requested weekday. Daily and interval rotations (`weekday=None`) are unaffected.

### Tests

Added `test_weekday_rotation_first_rotation_on_requested_weekday`, parametrized over `"w1 at 13:00"`, `"tuesday at 13:00"`, `"monday at 13:00"` (the broken cases) and `"w3 at 11:00"` (a time-before-creation control that was already correct and must stay so). The three broken cases fail on `master` (a spurious empty file appears) and pass with the fix; the control is unchanged. The full `tests/test_filesink_rotation.py` suite stays green (137 passed, 9 skipped); `black` and `ruff` are clean; `CHANGELOG.rst` updated.

---

*Disclosure: this contribution was prepared with AI assistance under my direction; I reviewed and tested all of it before submitting.*

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@
 - Add support for template strings used as log messages (`#1397 <https://github.com/Delgan/loguru/issues/1397>`_, thanks `@TurtleOrangina <https://github.com/TurtleOrangina>`_).
 - Fix incorrect microsecond value when formatting the log timestamp using ``{time:x}`` (`#1440 <https://github.com/Delgan/loguru/issues/1440>`_).
 - Fix parsing of 12-hour rotation times without seconds (`#1474 <https://github.com/Delgan/loguru/issues/1474>`_, thanks `@c-tonneslan <https://github.com/c-tonneslan>`_).
+- Fix a ``"<weekday> at <time>"`` rotation firing its first rotation on the creation day instead of the requested weekday when the time of day was later than the creation time (`#1484 <https://github.com/Delgan/loguru/pull/1484>`_).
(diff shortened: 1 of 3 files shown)


Delgan:
Very good catch. Thanks for that.

PR #1509, opened 2026-08-29
Fix new exception raised in async generator athrow() bypassing catch()

### LINKED ISSUE #1503: Exceptions raised by an async generator while handling `athrow()` bypass `@logger.catch`
## Summary

Async generator functions decorated with `@logger.catch` are wrapped by `AsyncGenCatchWrapper`.

The wrapper applies the configured catcher when the generator is resumed through `asend()`, but `athrow()` is forwarded directly to the underlying async generator:

```python
async def asend(self, value):
    with catcher:
        try:
            return await self._gen.asend(value)
        except StopAsyncIteration:
            pass
    raise StopAsyncIteration

async def athrow(self, *args, **kwargs):
    return await self._gen.athrow(*args, **kwargs)
```

This creates a behavioral inconsistency when an exception is injected through `athrow()` and the generator handles that exception but then raises a different exception of its own.

In that case, the new exception raised from inside the decorated async generator bypasses `@logger.catch`: it is not logged and `onerror` is not called.

The same exception raised from the generator body when execution proceeds through `asend()` is handled by `@logger.catch` as expected.

## Background

Async generator support for `@logger.catch` was introduced in #1302 / #1303.

There is already a test covering `athrow()` which verifies that an exception injected into the generator and left unhandled continues to propagate to the caller.

That behavior is reasonable and this issue does not require changing it.

The uncovered case is different:

1. The caller injects exception `A` with `athrow(A)`.
2. The async generator catches `A`.
3. While handling `A`, the generator raises a new exception `B`.
4. Exception `B`, although raised by the decorated generator itself, bypasses `@logger.catch`.

## Reproduction

```python
import asyncio

from loguru import logger


logger.remove()

logged = []

logger.add(
    lambda message: logged.append(message.record["message"]),
    format="{message}",
    colorize=False,
)


async def main():
    @logger.catch(reraise=True)
    async def via_asend():
        yield 1
        raise RuntimeError("asend-boom")

    gen = via_asend()
    await gen.asend(None)

    try:
        await gen.asend(None)
    except RuntimeError:
        pass

    print("asend logged:", logged)
    logged.clear()

    @logger.catch(reraise=True)
    async def via_athrow():
        try:
            yield 1
        except TimeoutError:
            raise RuntimeError("athrow-boom")

    gen = via_athrow()
    await gen.asend(None)

    try:
        await gen.athrow(TimeoutError)
    except RuntimeError as error:
        print("athrow raised:", error)

    print("athrow logged:", logged)


asyncio.run(main())
```

## Actual behavior

The `RuntimeError` raised while execution proceeds through `asend()` is logged:

```text
asend logged: ["An error has been caught in function 'main', process 'MainProcess' (...), thread 'MainThread' (...):"]
```

The `RuntimeError` raised by the generator while handling the exception injected through `athrow()` is not logged:

```text
athrow raised: athrow-boom
athrow logged: []
```

With `reraise=True`, the exception correctly continues to reach the caller, but it has bypassed the logging behavior provided by `@logger.catch`.

The same bypass also means that `onerror` is not invoked for this path.

## Expected behavior

An exception explicitly injected by the caller through `athrow()` and left unhandled by the generator can continue to propagate as it does currently.

However, if the generator handles the injected exception and then raises a new exception while executing its own code, that new generator-raised exception should be processed by `@logger.catch`, consistently with exceptions raised when the generator is resumed through `asend()`.

In other words, these two cases should remain distinguishable:

```python
async def generator():
    yield

await generator().athrow(ValueError)
```

Here, the injected `ValueError` is simply unhandled by the generator and may continue to propagate.

By contrast:

```python
async def generator():
    try:
        yield
    except ValueError:
        raise RuntimeError("failure while handling injection")
```

the `RuntimeError` originates from execution inside the decorated generator and would be expected to go through `@logger.catch`.

## Why this matters

`athrow()` is not limited to callers invoking the method manually.

It is part of Python's async-generator exception-injection protocol and is used by abstractions such as `contextlib.asynccontextmanager` when propagating exceptions from an `async with` body back into the underlying async generator.

This means the issue can occur in normal control-flow and cleanup code.

For example, conceptually:

```python
from contextlib import asynccontextmanager


@asynccontextmanager
@logger.catch(reraise=True)
async def resource():
    try:
        yield
    except SomeError:
        # Cleanup or error handling fails.
        raise RuntimeError("cleanup failed")
```

If the context manager implementation resumes the generator by injecting the original exception, a subsequent exception raised by the generator can follow the `athrow()` path and bypass the catcher.

## Existing test coverage

The existing async-generator `athrow()` test verifies the propagation of an unhandled injected exception.

It does not appear to cover the case where the generator catches the injected exception and raises a different exception while handling it.

A regression test for that distinction would make the intended semantics explicit.

## Possible direction

Simply wrapping all of `athrow()` in the same catcher used by `asend()` may alter the existing semantics for an injected exception that the generator does not handle, because the catcher could consume that injected exception when `reraise=False`.

Therefore, the main issue is not necessarily that `athrow()` should be wrapped identically to `asend()`.

Rather, the desired behavior would be to preserve propagation of an unhandled injected exception while ensuring that a new exception raised by the generator during `athrow()`-driven execution is still processed by `@logger.catch`.

The exact implementation is left open, since distinguishing these cases without changing the established `athrow()` behavior may require special handling.

## Environment

Reproduced with:

* Delgan/loguru commit `54c89d609fc98a63b5bf7f801119c5bc526e2016`
* CPython 3.11.2
* Linux x86_64

afonsojanu:
Closes #1503.

When a generator decorated with `logger.catch()` is an async generator, exceptions raised while it's resumed through `asend()` go through the configured catcher (logged, `onerror` called, optionally reraised). `athrow()` never had any of that: it just forwarded straight to the wrapped generator.

That's fine when the exception injected through `athrow()` propagates back out unhandled, since that's expected control flow rather than an error the decorator should be catching. But if the generator catches that injected exception and then raises a *different* exception while handling it, that new exception should go through the same catcher path as it would for `asend()`. Right now it doesn't, so it's silently never logged.

`athrow()` now normalizes whatever was passed in (class, class + value, or an instance, matching `athrow`'s own accepted forms) into a concrete exception instance, forwards that to the underlying generator, and checks identity on the way out:

- Same exception instance comes back out unhandled: passed through raw, exactly like before.
- A different exception comes out: routed through the catcher, so it's logged / triggers `onerror` / respects `reraise`, same as `asend()`.
- `StopAsyncIteration` (generator exhausted while handling the injected exception): passed through raw, mirroring how `asend()` already treats it.

This matters beyond direct `athrow()` calls since it's also how `contextlib.asynccontextmanager` propagates exceptions from an `async with` body back into the generator, so the bypass could show up in ordinary cleanup code without anyone calling `athrow()` by hand.

Added a regression test alongside the existing async generator decorator tests covering the case where a caught injected exception is handled by raising something new, verifying it's actually logged. Ran the full test suite plus black/ruff on the touched files.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,7 @@
 `Unreleased`_
 =============
 
+- Fix exceptions raised by an async generator while handling an injected ``athrow()`` exception bypassing ``@logger.catch()`` (`#1503 <https://github.com/Delgan/loguru/issues/1503>`_).
 - Change the default log format to include the timezone offset since it produces less ambiguous logs (`#856 <https://github.com/Delgan/loguru/pull/856>`_, thanks `@tim-x-y-z <https://github.com/tim-x-y-z>`_).
 - Add new ``logger.reinstall()`` method to automatically set up the ``logger`` in spawned child processes (`#818 <https://github.com/Delgan/loguru/issues/818>`_, thanks `@monchin <https://github.com/monchin>`_).
 - Add support for template strings used as log messages (`#1397 <https://github.com/Delgan/loguru/issues/1397>`_, thanks `@TurtleOrangina <https://github.com/TurtleOrangina>`_).
--- a/loguru/_logger.py
+++ b/loguru/_logger.py
@@ -1358,7 +1358,39 @@ class Logger:
                             raise StopAsyncIteration
 
                         async def athrow(self, *args, **kwargs):
-                            return await self._gen.athrow(*args, **kwargs)
+                            injected = self._normalize_thrown(*args)
+                            unhandled = None
+                            with catcher:
+                                try:
+                                    return await self._gen.athrow(injected)
+                                except StopAsyncIteration:
+                                    pass
+                                except BaseException as error:
+                                    if error is injected:
+                                        unhandled = error
+                                    else:
+                                        raise
+                            if unhandled is not None:
+                                raise unhandled
+                            raise StopAsyncIteration
+
+                        @staticmethod
+                        def _normalize_thrown(*args):
+                            if not args:
+                                return None
+                            exception = args[0]
+                            if isinstance(exception, BaseException):
+                                return exception
+                            if isclass(exception) and issubclass(exception, BaseException):
+                                value = args[1] if len(args) > 1 else None
+                                if value is None:
+                                    return exception()
+                                if isinstance(value, exception):
+                                    return value
+                                if isinstance(value, tuple):
+                                    return exception(*value)
+                                return exception(value)
+                            return None
 
                     def catch_wrapper(*args, **kwargs):
                         gen = function(*args, **kwargs)
--- a/tests/exceptions/source/modern/decorate_async_generator.py
+++ b/tests/exceptions/source/modern/decorate_async_generator.py
@@ -101,11 +101,41 @@ def test_decorate_async_generator_then_async_throw():
     asyncio.run(coro())
 
 
+def test_decorate_async_generator_athrow_with_new_error_reraised():
+    logged = []
+    logger.remove()
+    logger.add(lambda message: logged.append(message.record["message"]), format="{message}", colorize=False)
+
+    @logger.catch(reraise=True)
+    async def generator():
+        try:
+            yield 1
+        except TimeoutError:
+            raise RuntimeError("boom")
+
+    async def coro():
+        gen = generator()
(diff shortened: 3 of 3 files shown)


afonsojanu:
No worries if this doesn't fit right now, just let me know if there's anything you'd like adjusted, happy to revisit.

